PR #3, opened 2020-05-08
Batch scheduling

jterry64:


thomas-maschler:
Please upgrade gfw-terraform-module for fargate-autoscheduler to v0.1.1
This should fix the issue with dublicated Load Balancer name

app/tasks/batch.py
```diff
@@ -1,29 +1,30 @@
 from typing import Any, Dict, List, Optional, Callable, Awaitable, Set
 from time import sleep
 from datetime import datetime
+import os
 
 from app.models.pydantic.job import Job
 
+POLL_WAIT_TIME = 30
 BATCH_CLIENT = None
+REGION = os.environ.get("REGION", "us-east-1")
```
thomas-maschler, line 10: I created a module app/settings/global.py where we can import all the env variables. You can then import then from there

form ..settings.globale import REGION

thomas-maschler:
Left some comment for future reference. Will merge

PR #4, opened 2020-05-22
Feature/terraform batch

thomas-maschler:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no api changes)
- [X] Build related changes
- [x] Documentation content changes

Issue Number: N/A
- [GTC-507]
- [GTC-508]

## What is the new behavior?

- Add Terraform resources
- Add Vector pipeline
- Refactor code, pull CRUD (create, read, update, delete) actions into seperate module
- Build out test coverage

- [ ] Yes
- [x] No

## Other information

Vector pipeline is still redamental. We currently can only import standalone files, not composed files in ZIP archive (like SHP or GDB). However, implemenation of this should be straight forward. GDAL has a /vszip/ driver, which can unzip while reading. We just need to add the prefix to such files and the current scripts should work.


app/routes/__init__.py
```diff
@@ -26,27 +21,14 @@ async def version_dependency(
...
-        input_data["metadata"] = metadata
-
-    # new_row = model.from_orm(row)
-    # new_row.metadata = metadata
+async def version_dependency_form(
+    version: str = Form(..., title="Dataset version", regex=VERSION_REGEX)
+):
 
-    await row.update(**input_data).apply()
+    # if version == "latest":
```
jterry64, line 28: Is this something you meant to fill out later?

app/routes/analysis.py
```diff
@@ -0,0 +1,53 @@
+from enum import Enum
```
jterry64, line 1: did you mean to merge this in now? Probably should wait until it's actually working code, I'll just copy into a new a branch when I work on it

app/routes/versions.py
```diff
@@ -75,34 +76,84 @@ async def add_new_version(
     """
 
     async def callback(message: Dict[str, Any]) -> None:
-        await _version_history(message, dataset, version)
-
-    input_data, file_obj, uri = _prepare_sources(dataset, version, request, files)
+        pass
```
jterry64, line 79: Should you just get rid of this if you're not using it anymore?

app/routes/versions.py
```diff
@@ -223,35 +238,32 @@ async def _version_response(
...
-    else:
-        file_obj = None
-        uri = None
-
-    return input_data, file_obj, uri
-
-
-def _true_xor(*args):
-    return sum(args) == 1
+#
```
jterry64, line 241: can you delete this or will you need it later?

app/settings/globals.py
```diff
@@ -81,3 +80,22 @@
     port=WRITER_PORT,
     database=WRITER_DBNAME,
 )
+
+AWS_REGION = config("AWS_REGION", cast=str, default="us-east-1")
+
+POSTGRESQL_CLIENT_JOB_DEFINITION = config(
```
jterry64, line 86: should these have default values like this? If they're accidentally missing from the environment, we won't know until we get an error for trying to submit a batch job with a non-existent queue or job definition.

app/tasks/batch.py
```diff
@@ -1,31 +1,37 @@
...
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Set
+
+from ..models.pydantic.change_log import ChangeLog
+from ..models.pydantic.jobs import Job
+from ..settings.globals import POLL_WAIT_TIME
+from ..utils.aws import get_batch_client
+
+
+async def execute(
+    jobs: List[Job], callback: Callable[[Dict[str, Any]], Awaitable[None]]
```
jterry64, line 13: I think I'm the one who put this in here based on what you were doing in a different class, but should we actually use callbacks like this? It's being used specifically to add to the change log, and wouldn't work for any other function really. It should probably be more clear in this is just a change logger.

app/tasks/default_assets.py
```diff
@@ -0,0 +1,80 @@
...
+
+    async with ContextEngine("PUT"):
+        await versions.update_version(
+            dataset, version, status=status, change_log=[log.dict()]
+        )
+
+
+async def _inject_file(file_obj: IO, s3_uri: str) -> ChangeLog:
+    """
+    Upload a file-like object to S3 data lake
```
jterry64, line 63: Do you think we'll need to worry about that eventual consistency issue again with S3? (where uploaded files may not appear for all readers for ~15 min)

app/tasks/table_source_assets.py
```diff
@@ -0,0 +1,99 @@
...
+    new_asset = await assets.create_asset(**data.dict())
+
+    create_table_job = PostgresqlClientJob(
+        job_name="create table", command=["create table"], environment=writer_secrets,
+    )
+
+    load_data_jobs: List[Job] = list()
+
+    for i, uri in enumerate(source_uris):
+        load_data_jobs.append(
```
jterry64, line 40: These can all write in parallel ok? Do we need to set a limit on the number of parallel writers, or will Aurora automatically handle them properly?

app/tasks/table_source_assets.py
```diff
@@ -0,0 +1,99 @@
...
+                command=["build_index"],
+                environment=writer_secrets,
+                parents=[gfw_attribute_job.job_name],
+            )
+        )
+
+    cluster_jobs: List[Job] = list()
+    if config_options.partitions:
+        for partition in config_options.partitions:
+            index_jobs.append(
```
jterry64, line 71: Is this correct? It looks like you're just doing the same command as the index job above, instead of partitioning

app/tasks/table_source_assets.py
```diff
@@ -0,0 +1,99 @@
...
+        environment=writer_secrets,
+        parents=[job.job_name for job in load_data_jobs],
+    )
+
+    index_jobs: List[Job] = list()
+
+    for index in config_options.indices:
+        index_jobs.append(
+            PostgresqlClientJob(
+                job_name="geom index",
```
jterry64, line 61: since this is just tabular, would there be a geom index?

app/tasks/vector_source_assets.py
```diff
@@ -0,0 +1,150 @@
...
+                    index.column_name,
+                    "-x",
+                    index.index_type,
+                ],
+                parents=[gfw_attribute_job.job_name],
+                environment=writer_secrets,
+            )
+        )
+
+    inherit_geostore_job = PostgresqlClientJob(
```
jterry64, line 121: Should this pipeline also check for partitions? I thought you partition the fire vector data

app/utils/aws.py
```diff
@@ -0,0 +1,45 @@
...
+        return service_client
+
+    return client
+
+
+get_s3_client = client_constructor("s3")
+get_batch_client = client_constructor("batch")
+
+#
+# def get_s3_client():
```
jterry64, line 30: Delete commented out code if no longer needed

terraform/main.tf
```diff
@@ -16,42 +16,132 @@ provider "aws" {
...
+  tags                  = data.terraform_remote_state.core.outputs.tags
+  name_suffix           = terraform.workspace == "default" ? "" : "-${terraform.workspace}"
+  project               = "gfw-data-api"
+  aurora_instance_class = data.terraform_remote_state.core.outputs.aurora_cluster_instance_class
+  aurora_max_vcpus      = local.aurora_instance_class == "db.t3.medium" ? 2 : local.aurora_instance_class == "db.r4.large" ? 2 : local.aurora_instance_class == "db.r4.xlarge" ? 4 : local.aurora_instance_class == "db.r4.2xlarge" ? 8 : local.aurora_instance_class == "db.r4.4xlarge" ? 16 : local.aurora_instance_class == "db.r4.8xlarge" ? 32 : local.aurora_instance_class == "db.r4.16xlarge" ? 64 : local.aurora_instance_class == "db.r5.large" ? 2 : local.aurora_instance_class == "db.r5.xlarge" ? 4 : local.aurora_instance_class == "db.r5.2xlarge" ? 8 : local.aurora_instance_class == "db.r5.4xlarge" ? 16 : local.aurora_instance_class == "db.r5.8xlarge" ? 32 : local.aurora_instance_class == "db.r5.12xlarge" ? 48 : local.aurora_instance_class == "db.r5.16xlarge" ? 64 : local.aurora_instance_class == "db.r5.24xlarge" ? 96 : ""
```
jterry64, line 25: is there no easier way in terraform to write out this logic haha?

thomas-maschler: Ha, I wish!
I haven't found a way to expose/ extract the CPU count of the RDS instance type. If you have a better solution, please let me know

app/routes/__init__.py
thomas-maschler: I reviewed this again and I prefer the 307 redirect approach. This is cleaner, since the client will know which latest version was received.

app/tasks/batch.py
jterry64: maybe just pass some kind of StatusLogger object, which accepts the basic fields (status/message/detail) and the caller will instantiate object with whatever other details are needed to log correctly (e.g. dataset/version)?

thomas-maschler: We should probably make this its own feature. I will create a JIRA ticket for this and cover this in a different PR

PR #25, opened 2020-07-02
Chunk source URIs when creating assets

jterry64:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes

Each source URI when creating a table asset is load to the db in a parallel batch job. These batch jobs are then the dependencies for other batch jobs in the pipeline. However, AWS Batch has a limit of 20 jobs as dependencies, so this break if there are more than 20 source URIs.

## What is the new behavior?
Chunk the source URIs so that there's never more than 20 chunks. These chunks will get loaded to the db in the same batch job.

- [ ] Yes
- [X] No

## Notes

I think this fixes the issue. It should still run jobs in parallel if there's <20 source URIs, but if there's more, it'll break than down into parallel chunks that stay in the limit. I made a new test that's just the same as test_table_source_asset, but I mock upload 99 files to S3, one for each row in test.tsv (so it's not super slow). This seems to work, let me know if it doesn't seem like a valid test.

### DIFF
--- a/app/tasks/batch.py
+++ b/app/tasks/batch.py
@@ -10,6 +10,8 @@ from ..models.pydantic.change_log import ChangeLog
 from ..models.pydantic.jobs import Job
 from ..utils.aws import get_batch_client
 
+BATCH_DEPENDENCY_LIMIT = 20
+
 
 async def execute(jobs: List[Job],) -> ChangeLog:
     try:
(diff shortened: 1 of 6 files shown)


thomas-maschler:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- app/tasks/table_source_assets.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/wri/gfw-data-api/pullRequest?prid=5795644&bid=18969331)

PR #88, opened 2020-09-16
Async lambda calls

jterry64:
Added an async boto3 client to make sure lambda calls for raster analysis are done asynchronously, since the API may have to wait 15-20 seconds for a response. Without the async boto3 client, concurrent user requests won't send lambda requests concurrently. 

Also added a POST endpoint for analysis calls to directly send the geometry (needed this for stress testing, figured it'd be generally useful).

## Pull request checklist

- [ X Make sure you are making a pull request against the develop branch (left side). Also you should start your branch off our develop.

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

The lambda request is synchronously, so multiple requests at the same time will hold up a worker thread until it returns. 

## What is the new behavior?

- Concurrent user requests for raster analysis will actually run concurrently
- POST requests can be sent to API directly with the geometry

- [ ] Yes
- [X] No

## Other information

aiboto3 and moto don't place nice with each other right now, so I had to monkeypatch the lambda invoke for tests to use the normal boto3 client. There's a bug for it on github, I can remove this once it's resolved: https://github.com/aio-libs/aiobotocore/issues/755


thomas-maschler:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 2
- Added 6
           

Complexity increasing per file
==============================
- app/models/pydantic/analysis.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/wri/gfw-data-api/pullRequest?prid=6190258&bid=20198153)

Pipfile
```diff
@@ -38,6 +38,8 @@ pglast = "*"
 async-lru = "*"
 httpx = "*"
 pendulum = "*"
+aioboto3 = "*"
+typing-extensions = "==3.7.4.3"
```
thomas-maschler, line 42: Do you really need this?
Sounds like this is a backport of typing for python 3.5/ 3.6

_Users of other Python versions should continue to install and use the typing module from PyPi instead of using this one unless specifically writing code that must be compatible with multiple Python versions or requires experimental types._

app/routes/analysis/analysis.py
```diff
@@ -12,8 +12,9 @@
 
 from ...models.enum.geostore import GeostoreOrigin
 from ...models.pydantic.responses import Response
-from app.settings.globals import RASTER_ANALYSIS_LAMBDA_NAME
-from app.utils.aws import get_lambda_client
+from ...models.pydantic.analysis import ZonalAnalysisRequestIn
+from app.settings.globals import RASTER_ANALYSIS_LAMBDA_NAME, AWS_REGION
+import aioboto3
```
thomas-maschler, line 17: make path relative, sort import statements using isort

app/routes/analysis/analysis.py
```diff
@@ -38,7 +39,43 @@ async def zonal_statistics(
...
+        request.start_date,
+        request.end_date,
+    )
+
+
+async def _zonal_statics(
+    geometry: Dict[str, Any],
+    sum_layers: List[RasterLayer],
+    group_by: Optional[List[RasterLayer]],
+    filters: Optional[List[RasterLayer]],
```
thomas-maschler, line 75: Looking at the signature above again (line 35 and 36).

If default values for group_by and filters is `[]` type is no longer `Optional`. It should be `None`. When using value `[]` you can simply use `List[RasterLayer]`

app/routes/analysis/analysis.py
```diff
@@ -48,16 +85,22 @@ async def zonal_statistics(
...
-    response_payload = json.loads(response['Payload'].read().decode())
+    response_payload_encoded = await _invoke_lambda(payload)
+    response_payload = json.loads(response_payload_encoded.decode())
 
     if response_payload['statusCode'] != 200:
         raise InvalidResponseError(f"Raster analysis returned status code {response_payload['statusCode']}")
 
     response_data = response_payload['body']['data']
     return Response(data=response_data)
+
```
thomas-maschler, line 96: In line 92 raise `HTTPException` with status code 400 or 500 instead (user/ server error) and add a more meaningful error message. End user will see this error when making a call.

tests/routes/test_analysis.py
```diff
@@ -1,9 +1,37 @@
...
+    async def mock_invoke(payload):
+        response = get_lambda_client().invoke(
+            FunctionName=RASTER_ANALYSIS_LAMBDA_NAME,
+            InvocationType="RequestResponse",
+            Payload=bytes(json.dumps(payload), "utf-8"),
+        )
+
+        return response['Payload'].read()
+
+    monkeypatch.setattr(analysis, "_invoke_lambda", mock_invoke)
```
thomas-maschler, line 34: Did you look into the moto server option?
Hope we find some time at one point to investigate localstack as an alternative.

thomas-maschler (changes requested):

Pipfile
jterry64: It was needed for aiboto3 to install properly for some reason, otherwise it would give complaints that it couldn't find it. Their PyPi says they support 3.8, but I guess still have dependencies on this library.

tests/routes/test_analysis.py
jterry64: Yeah, I couldn't make the lambda server work correctly because it depended on shared state with the IAM service (for assuming the lambda role), but you can't run multiple services on the same moto server. Localstack does support that, and that's what people online recommended. It didn't seem worth it to move everything to Localstack just for this small change, but I'm in favor of looking into this soon

jterry64: Actually I'll just quickly look now if I can set it up just for this test suite, and we can migrate the rest of the tests later

jterry64: Hmm, for some reason aioboto3 only runs on a very pretty version of boto3 (1.12.32) which localstack doesn't support (>=1.14.33). So still might not work here, although localstack is pretty easy to set up, so we should definitely put some time in to switch over soon

thomas-maschler: ok, thanks for investigating.
Let's work with Liza to put some time aside to improve the test setup.

thomas-maschler: Just another thought. What happens if you only use aiobotocore? 
There would be less dependencies on external libraries. Would you still have botocore version restrictions?
Or do you need boto3 to use moto server/ localstack?

thomas-maschler:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 2
- Added 6
           

Complexity increasing per file
==============================
- app/models/pydantic/analysis.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/wri/gfw-data-api/pullRequest?prid=6190258&bid=20198153)

tests/routes/test_analysis.py
jterry64: That actually works if I downgrade localstack a little. Although I realized I really only need to run it in a docker container, the library itself is in unnecessary on the server. But I still can't seem to get it to work, it just keeps giving me weird and undescriptive errors when it runs my test lambda that moto had no issue with it. 

If I can't figure it out by Monday, can we just go with this test and move to localstack when we set it all up? I'd like to get OTF moving towards production.

thomas-maschler: Thanks for looking into this. And yes, I am fine with putting this on the backburner.

PR #103, opened 2020-11-23
GTC-813 Implement raster tile cache for RADD/GLAD

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Raster tile cache not implemented
Issue Number: N/A

## What is the new behavior?

- Allows creation of raster tile caches for the RADD/GLAD use case
-
-

- [ ] Yes
- [x] No


app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+    parents: List[Job],
+):
+    merge_intensity_jobs: List[Job] = []
+
+    for zoom_level in range(min_zoom, max_zoom):
+        # Sanitize creation_options
+
+        d_c_co = copy.deepcopy(date_conf_co)
+        d_c_co["srid"] = "epsg-3857"
+        d_c_co["grid"] = f"zoom_{zoom_level}"
```
thomas-maschler, line 311: You don't use `date_conf_co` and `intensity_co` elsewhere. You don't need to use deepcopy, do you?
since srid is not part of the model, i would keep it seperate. In make it a separate argument in `get_asset_uri`  and have it default to `epsg:4326`.
Please note that the actual srid notation has a colon `:` not a dash `-`. I replaced it with the dash b/c otherwise it would conflict with standard path naming conventions. In `get_asset_uri` you will want to sanitize the srid argument and replace colon with dash

app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+        asset_options = AssetCreateIn(
+            asset_type=AssetType.raster_tile_set,
+            asset_uri=asset_uri,
+            is_managed=True,
+            creation_options=co_obj,
+            metadata={},
+        ).dict(by_alias=True)
+
+        wm_asset_record = await create_asset(dataset, version, **asset_options)
+        print(
```
thomas-maschler, line 349: use `logger`

app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+        intensity_uri = get_asset_uri(
+            dataset, version, AssetType.raster_tile_set, i_co
+        ).replace("{tile_id}.tif", "tiles.geojson")
+
+        c_co = copy.deepcopy(intensity_co)
+        c_co["srid"] = "epsg-3857"
+        c_co["grid"] = f"zoom_{zoom_level}"
+        c_co["pixel_meaning"] = RGB_ENCODED_PIXEL_MEANING
+        asset_uri = get_asset_uri(dataset, version, AssetType.raster_tile_set, c_co)
+        merged_asset_prefix = asset_uri.rsplit("/", 1)[0]
```
thomas-maschler, line 328: This is the same as os.path.dirname right? I believe the this would be easier to read, don't you think?

app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+
+        co = copy.deepcopy(r_t_s_creation_options)
+        co["grid"] = f"zoom_{zoom_level}"
+        asset_prefix = get_asset_uri(
+            dataset, version, AssetType.raster_tile_set, co
+        ).rsplit("/", 1)[0]
+
+        print(
+            f"CREATING TILE CACHE JOB FOR ZOOM LEVEL {zoom_level} WITH PREFIX {asset_prefix}"
+        )
```
thomas-maschler, line 397: Use `logger`

app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+    tile_cache_jobs: List[Job] = []
+
+    for zoom_level in range(min_zoom, max_zoom):
+        # Sanitize creation_options
+
+        co = copy.deepcopy(r_t_s_creation_options)
+        co["grid"] = f"zoom_{zoom_level}"
+        asset_prefix = get_asset_uri(
+            dataset, version, AssetType.raster_tile_set, co
+        ).rsplit("/", 1)[0]
```
thomas-maschler, line 393: see above

batch/pixetl.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM globalforestwatch/pixetl:v1.0.0
+FROM globalforestwatch/pixetl:v1.1.0
```
thomas-maschler, line 1:
```suggestion
FROM globalforestwatch/pixetl:v1.2.0
```

tests/conftest.py
```diff
@@ -148,6 +149,24 @@ def client():
     with TestClient(app) as client:
         yield client
 
+        # Clean up created assets/versions/datasets so teardown doesn't break
```
thomas-maschler, line 152: Why does it break?

tests/conftest.py
```diff
@@ -5,6 +5,7 @@
 import threading
 import zipfile
 from http.server import HTTPServer
+from time import sleep
```
thomas-maschler, line 8: why sleep?

batch/python/merge_intensity.py
```diff
@@ -0,0 +1,123 @@
...
+            print(f"No intensity raster file for coordinates {k}... Skipping")
+            continue
+        else:
+            output_uri = "/".join([destination_prefix, intensity_uri.rsplit("/", 1)[1]])
+            print("About to operate on:")
+            print(f"date_conf file: {date_conf_uri}")
+            print(f"intensity file: {intensity_uri}")
+            print(f"output file: {output_uri}")
+
+            process_rasters(date_conf_uri, intensity_uri, output_uri)
```
thomas-maschler, line 79: I am loosing concentration, but this seems overly complicated. 
file names should be always the same for matching tiles. so you could do something like this
```python
def get_tile_ids(bucket, key):
    response = s3_client.get_object(Bucket=bucket, Key=key)
    geojson: dict = json.loads(response["Body"].read().decode("utf-8"))    
    tiles = [os.path.basename(feature["properties"]["name"]) for feature in geojson["features"]]
    return tiles

date_conf_tiles = get_tile_ids(bucket, date_conf_key)
intensity_tiles = get_tile_ids(bucket, intensity_key)

common_tiles = set(date_conf_tiles) & set(intensity_tiles)

for tile in common_tiles:
    date_conf_tile = os.path.join(os.path.basedir(date_conf_uri), tile)
    intensity_tile = os.path.join(os.path.basedir(intensity_uri), tile)
    output_tile = os.path.join(output_prefix, tile)
    process_rasters(date_conf_tile, intensity_tile, output_tile)
```

batch/python/merge_intensity.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def process_rasters(date_conf_uri, intensity_uri, output_uri):
+    s3_client = get_s3_client()
+
+    # Download both files into a temporary directory
+    with TemporaryDirectory() as temp_dir:
+        local_date_conf_path = os.path.join(temp_dir, "date_conf.tiff")
+        local_intensity_path = os.path.join(temp_dir, "intensity.tiff")
+        local_output_path = os.path.join(temp_dir, "output.tiff")
```
thomas-maschler, line 89: should be `.tif`

batch/python/raster_tile_cache.py
```diff
@@ -0,0 +1,117 @@
...
+        print(proc.stdout)
+        print(proc.stderr)
+        proc.check_returncode()
+
+        cmd_arg_list = [
+            "gdal2tiles.py",
+            f"--zoom={zoom_level}",
+            "--s_srs",
+            "EPSG:3857",
+            "--resampling=bilinear",
```
thomas-maschler, line 81: we do only one zoom level at a time and input zoom level is identical to output zoom level . use `resampling=nearest`

batch/python/raster_tile_cache.py
```diff
@@ -0,0 +1,117 @@
...
+            "--processes=1",  # FIXME: Pass in cores value?
```
thomas-maschler, line 82: We want to us as many cores as possible!
```suggestion
            f"--processes=os.environ.get('CORES', os.cpu_count())",  # FIXME: Pass in cores value?
```

batch/python/raster_tile_cache.py
```diff
@@ -0,0 +1,117 @@
...
+            implementation,
+            tiles_dir,
+        ]
+        print(f"Running command: {cmd_arg_list}")
+        proc: subprocess.CompletedProcess = subprocess.run(
+            cmd_arg_list, capture_output=True, check=False, text=True
+        )
+        print(proc.stdout)
+        print(proc.stderr)
+        proc.check_returncode()
```
thomas-maschler, line 113: you can also use the python module directly.

```python
from tileputty.upload_tiles import upload_tiles

upload_tiles(
    tiles_dir,
    dataset,
    version,
    bucket=target_bucket,
    implementation=implementation,
)
```

thomas-maschler (changes requested):

app/tasks/raster_tile_cache_assets.py
```diff
@@ -0,0 +1,423 @@
...
+    max_zoom = input_data["creation_options"]["max_zoom"]
+    max_static_zoom = input_data["creation_options"]["max_static_zoom"]
+    implementation = input_data["creation_options"]["implementation"]
+
+    assert min_zoom <= max_zoom  # FIXME: Raise appropriate exception
+    assert max_static_zoom <= max_zoom  # FIXME: Raise appropriate exception
+
+    # FIXME: Remove this when implementing standard tile cache code path:
+    if (
+        input_data["creation_options"]["symbology"]["color_map"]["type"]
```
thomas-maschler, line 64:
```suggestion
        input_data["creation_options"]["symbology"]["type"]
```

thomas-maschler (changes requested):

tests/conftest.py
dmannarino: Because during teardown it downgrades all of the migrations during which it attempts to re-instate a unique constraint involving (if I recall correctly) dataset, version, and assettype. Since by the end of some tests we have multiple assets of the same type (in particular raster tile sets), this breaks teardown, leaving a mess. That's why I occasionally mention the possibility of squashing migrations. I initially a implemented something like this at the end of all the RTS/RTC tests, but I decided it was better to clean things out more generally between runs. This also means we don't have dataset collisions between tests if it makes sense to reuse the same names.

thomas-maschler: ok, +1 to squash migrations

PR #112, opened 2020-12-23
Save stats, extent to raster tile set assets

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- Stats endpoints exist but records have no stats saved
- Extent endpoint/saving is not implemented

Issue Number: GTC-796, GTC-956

## What is the new behavior?

- stats and extent are fully functional for raster tile sets

- [ ] Yes
- [x] No


tests/routes/test_assets.py
```diff
@@ -519,3 +520,219 @@ def _delete_s3_files(bucket, prefix):
...
+    asset_resp = await async_client.get(f"/asset/{asset_id}/stats")
+    version_resp = await async_client.get(f"/dataset/{dataset}/{version}/stats")
+
+    for resp in (asset_resp, version_resp):
+        assert resp.json()["data"]["bands"][0]["min"] == 1.0
+        assert resp.json()["data"]["bands"][0]["max"] == 1.0
+        assert resp.json()["data"]["bands"][0]["mean"] == 1.0
+        assert resp.json()["data"]["bands"][0]["histogram"]["bin_count"] == 256
+        assert resp.json()["data"]["bands"][0]["histogram"]["value_count"][255] == 0
+        # FIXME: I think the following assertion should be true, but the value always
```
jterry64, line 572: did you ever figure this out?

app/models/pydantic/statistics.py
```diff
@@ -77,33 +80,41 @@ class Affine(BaseModel):
...
     min: float = Field(..., description="Minimum pixel value.")
     max: float = Field(..., description="Maximum pixel value.")
     mean: float = Field(..., description="Mean pixel value.")
-    std_dev: float = Field(..., description="Standard deviation of pixel values.")
-    histogram: Histogram = Field(..., description="Histogram.")
+
+    histogram: Optional[Histogram] = Field(description="Histogram.")
+
+    class Config:
+        extra = "forbid"
```
jterry64, line 109: I'm a little confused what all these extra = "forbid" attributes are for

jterry64 (changes requested):

app/models/pydantic/statistics.py
dmannarino: Putting that in a model means that you can't instantiate the model with "extra" fields. So if the model is
class Blah(BaseModel):
    foo: str
    class Config:
        extra = "forbid"
and you try to do
Blah(**{"foo": "something", "bar": "something else"})
an exception will be thrown because "bar" is not a valid field for the Blah model.
Without extra="forbid" the extra fields will be silently thrown away, which can lead to errors being missed and be really confusing when debugging.

app/models/pydantic/statistics.py
```diff
@@ -77,33 +80,41 @@ class Affine(BaseModel):
...
-    nodata: Optional[int] = Field(..., description="No data value.")
-    dtype: str = Field(..., description="Pixel data type.")
-    width: int = Field(..., description="Raster width.")
-    height: int = Field(..., description="Raster height.")
-    bounds: Tuple[float, float, float, float] = Field(..., description="Raster founds.")
-    transform: Affine = Field(..., description="Affine transformation.")
-    crs: str = Field(..., description="Coordinate reference system.")
-    pixel_size: Tuple[float, float] = Field(..., description="Raster pixel size.")
-    origin: Tuple[float, float] = Field(..., description="Raster origin.")
+    # driver: str = Field(..., description="Driver used to create raster file.")
```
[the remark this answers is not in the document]
dmannarino: Yeah. Thomas said to leave out the extra info for now, but I didn't want to completely erase it. Will insert a comment to that effect so people don't wonder.

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
```diff
@@ -55,3 +68,98 @@ async def raster_tile_set_asset(
...
+    bucket, tiles_key = split_s3_path(tile_uri_to_tiles_geojson(asset_uri))
+
+    s3_client = get_s3_client()
+    tiles_resp = s3_client.get_object(Bucket=bucket, Key=tiles_key)
+    tiles_geojson: dict = json.loads(tiles_resp["Body"].read().decode("utf-8"))
+
+    stats_by_band: DefaultDict[int, DefaultDict[str, List[int]]] = defaultdict(
+        lambda: defaultdict(lambda: [])
+    )
+    histogram_by_band: Dict[int, Dict[str, Any]] = dict()
```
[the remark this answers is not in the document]
dmannarino: Good idea. I was at least able to replace Dict[str, Any] with Histogram.

tests/routes/test_assets.py
dmannarino: Nope. Will Ask Thomas on Monday. I think I probably just don't understand how things work.

jterry64 (approved):

app/routes/tasks/task.py
```diff
@@ -180,24 +198,14 @@ async def _check_completed(asset_id: UUID):
...
-            asset_row = await _update_asset_field_metadata(
-                asset_row.dataset, asset_row.version, asset_id,
-            )
-
-            await _register_dynamic_vector_tile_cache(
-                asset_row.dataset, asset_row.version, asset_row.metadata
-            )
 
-        # If default asset, make sure, version is also set to saved
+        # If default asset, make sure version is also set to saved
```
thomas-maschler, line 208: Set version to default should run after `post_completion_task` in case that step fails

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
```diff
@@ -48,10 +63,116 @@ async def raster_tile_set_asset(
...
+):
+    if target_histo_i is None:
+        target_histo_i = Histogram(
+            **{
+                "min": source_histo_i["min"],
+                "max": source_histo_i["max"],
+                "bin_count": source_histo_i["count"],
+                "value_count": source_histo_i["buckets"],
+            }
+        )
```
thomas-maschler, line 104:
```suggestion
        target_histo_i = Histogram(
                min=source_histo_i["min"],
                max=source_histo_i["max"],
                bin_count=source_histo_i["count"],
                value_count=source_histo_i["buckets"]           
        )
```

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
```diff
@@ -48,10 +63,116 @@ async def raster_tile_set_asset(
...
+    else:
+        target_histo_i.min = min(target_histo_i.min, source_histo_i["min"])
+        target_histo_i.max = max(target_histo_i.max, source_histo_i["max"])
+        for bucket_num, bucket_val in enumerate(source_histo_i["buckets"]):
+            target_histo_i.value_count[bucket_num] += bucket_val
```
thomas-maschler, line 109: unfortunately this won't work all the time.

GDAL always returns a histogram with 256 buckets. Bucket size is defined by `min`, `max` and data type.

Ie if `min=0` and `max=1` and `datatype=int`, bucket size will be 1. Hence the first bucket will list the count for all `0`, the second bucket will list the counts for all `1` and the remaining 254 buckets will be empty.

if we have a `min=0` and `max=1000` and we have any values in between, it will be more tricky. In that case, bucket size will be most likely `4` and values will be grouped together in their corresponding bucket. Ie values `0-3` will be counted in the first bucket, values `4-7` in the second and so on.

If we now try to merge to histograms, we will have to make sure that the bucket size and the min/ max are the same, otherwise buckets won't line up. Ie, one raster might have a bucket size of 3 and a range of 200 - 900 and the other one a size of 4 and a range of 0 - 1000.

A possible solution on how to address this issue is described here
https://stackoverflow.com/a/47183002/1410317

tests/routes/test_assets.py
thomas-maschler: Please see my comments about histograms above

thomas-maschler (changes requested):

app/routes/assets/asset.py
```diff
@@ -200,22 +201,34 @@ async def get_change_log(asset_id: UUID = Path(...)):
...
     "/{asset_id}/stats",
     response_class=ORJSONResponse,
     tags=["Assets"],
     response_model=StatsResponse,
 )
 async def get_stats(asset_id: UUID = Path(...)):
-    asset = await assets.get_asset(asset_id)
-    stats: Optional[Stats] = stats_factory(asset.asset_type, **asset.stats)
+    asset: ORMAsset = await assets.get_asset(asset_id)
+    stats: Optional[Stats] = stats_factory(asset.asset_type, asset.stats)
```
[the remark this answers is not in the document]
dmannarino: We used to default to {}, I changed it to default to Null. Updated signature, thanks!

tests/routes/test_assets.py
dmannarino: Thomas figured it out: I was over-sampling the pixels in the source data.

app/models/pydantic/statistics.py
dmannarino: Done. Everything that was using BaseModel is now using DataApiBaseModel.

app/routes/tasks/task.py
dmannarino: Done.

PR #130, opened 2021-01-28
GTC-978: Guard against spot instance theft; Enable GFW geostores for analysis

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- Spot instances are frequently stolen back by AWS, and we don't deal with it gracefully. No attempt is made to retry, because we don't want to retry if an application exits abnormally.
- GFW geostores cannot be used with the raster analysis lambda (despite that being the API default), only RW geostores
- The use of aioboto3 requires us to use an ancient collection of *boto* modules

Issue Number: GTC-978

## What is the new behavior?

- Batch jobs will retry if their spot instance is stolen, but not if the application exits abnormally
- GFW geostores can be used with the raster analysis lambda
- The use of aioboto3 is replaced with an async HTTP call, allowing use of modern *boto* releases

- [ ] Yes
- [x] No

## Other information

Unfortunately the lambda change is not compatible with moto, so I had to skip the analysis tests for now (until we switch to localstack). This should therefore be tested manually in staging.


app/tasks/batch.py
```diff
@@ -122,25 +122,15 @@ def submit_batch_job(
             "memory": job.memory,
             "environment": job.environment,
         },
-        "retryStrategy": {"attempts": job.attempts},
+        "retryStrategy": {
+            "attempts": job.attempts,
+            "evaluateOnExit": [{"onStatusReason": "Host EC2*", "action": "RETRY"}],
```
thomas-maschler, line 127: Nice, this gives me the idea that we could utilize a different exit code (ie 2) in PixETL for when some of the tiles failed but not all. We could leverage this to trigger a retry. In that case we could add another `evaluateOnExit` with `onExitCode`

app/utils/geostore.py
```diff
@@ -1,25 +1,40 @@
...
+        GeostoreOrigin.rw: rw_api.get_geostore_geometry,
     }
 
     try:
         return await geostore_constructor[geostore_origin](geostore_id)
-    except KeyError:
-        raise HTTPException(
-            status_code=501,
-            detail=f"Geostore origin {geostore_origin} not fully implemented.",
-        )
```
thomas-maschler, line 22: I believe you want to keep this, incase we add another geostore origin at a later point. This will help us to narrow down the issue.

app/utils/geostore.py
```diff
@@ -1,25 +1,40 @@
...
 
+@alru_cache(maxsize=128)
+async def _get_gfw_geostore_geometry(geostore_id: UUID) -> Geometry:
+    """Get GFW Geostore geometry."""
+
+    geo: GeostoreHydrated = await get_geostore_from_anywhere(geostore_id)
+
+    try:
+        geometry = geo.gfw_geojson.features[0].geometry
+    except KeyError:
```
thomas-maschler, line 21: Also need to catch `RecordNotFoundError` error and raise a `BadResponseError`, if geostore_id does not return a match

app/routes/analysis/analysis.py
```diff
@@ -93,27 +102,46 @@ async def _zonal_statics(
...
-    response_payload = json.loads(response_payload_encoded.decode())
+    try:
+        response = await _invoke_lambda(payload)
+    except httpx.TimeoutException:
+        raise HTTPException(500, "Query took too long to process.")
 
-    if response_payload["statusCode"] != 200:
+    resp_dict = response.json()
+
+    if response.status_code != 200 or resp_dict.get("statusCode") != 200:
```
thomas-maschler, line 112: While it looks like Justin always returns a json like objection, there might be a case where the Lambda function fails and does not return a json like object but a string.

how about 
```python
try:
    response_data = response.json()["body"]["data"]
except KeyError, <HTTPX error for when response is not json-serializable>:
    logger.error(
           f"Raster analysis lambda experienced an error. Full response: {response.text}"
        )	        
       raise HTTPException(500, "Raster analysis geoprocessor experienced an error. See logs."	)

return Response(data=response_data)
```

app/routes/analysis/analysis.py
```diff
@@ -93,27 +102,46 @@ async def _zonal_statics(
...
         logger.error(
-            f"Raster analysis lambda returned status code {response_payload['statusCode']}"
+            f"Raster analysis lambda experienced an error. Full response: {resp_dict}"
         )
         raise HTTPException(
             500, "Raster analysis geoprocessor experienced an error. See logs."
         )
 
-    response_data = response_payload["body"]["data"]
+    response_data = response.json()
```
thomas-maschler, line 120: https://github.com/wri/gfw-raster-analysis-lambda/blob/master/lambdas/tiled_analysis/src/lambda_function.py#L44

you only want to return `response.json()["body"]["data"]`. See comment above

thomas-maschler (changes requested):

app/routes/analysis/analysis.py
thomas-maschler: https://github.com/encode/httpx/blob/master/httpx/_models.py#L1110-L1118

They are just using json.loads so error would be `json.decoder.JSONDecodeError`

app/tasks/batch.py
dmannarino: Yup! You can add up to five retry conditions.

app/utils/geostore.py
dmannarino: Okay, sure.

thomas-maschler (approved):

dmannarino:
Obsoleted by Thomas's PR, which cherry-picked these changes

PR #141, opened 2021-02-16
update version regex to support v1.8 schema

thomas-maschler:
(cherry picked from commit 5d15f9dcc9fb218ebaa02d9a64dff7eaf8a0ccc5)

## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Fix version regex
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/routes/__init__.py
+++ b/app/routes/__init__.py
@@ -10,7 +10,7 @@ from ..errors import RecordNotFoundError
 from ..utils.rw_api import who_am_i
 
 DATASET_REGEX = r"^[a-z][a-z0-9_-]{2,}$"
-VERSION_REGEX = r"^v\d{1,8}\.?\d{1,3}\.?\d{1,3}$|^latest$"
+VERSION_REGEX = r"^v\d{1,8}(\.\d{1,3}){0,2}?$|^latest$"
 oauth2_scheme = OAuth2PasswordBearer(tokenUrl="/token")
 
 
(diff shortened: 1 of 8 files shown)


app/routes/__init__.py
```diff
@@ -10,7 +10,7 @@
 from ..utils.rw_api import who_am_i
 
 DATASET_REGEX = r"^[a-z][a-z0-9_-]{2,}$"
-VERSION_REGEX = r"^v\d{1,8}\.?\d{1,3}\.?\d{1,3}$|^latest$"
+VERSION_REGEX = r"^v\d{1,8}(\.\d{1,3}){0,2}?$|^latest$"
```
jterry64, line 13: Is there a reason we actually need to enforce version names like this? E.g. for data updates, it'd be nice to have a version like v2020Q1. Why not just allow any alphanumeric character or period with a size limit?

jterry64 (approved):

app/routes/__init__.py
thomas-maschler: Good question.

I prefer to have versioning schemas which are sortable so that the release order is clear. If allowing any alphanumeric character this could grow quickly out of control. Instead you could do `v2020.1`

jterry64: Ah ok, that makes sense to me.

PR #148, opened 2021-02-24
Feature/multisource

thomas-maschler:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

## What is the new behavior?

- allow possibility to have multiple raster input sources
- You can specify multiple input sources when creating a new version
- You can specify additional auxiliary assets when creating a new asset. In this case, default asset and auxiliary assets will be used as sources

- [ ] Yes
- [x] No

## Other information

Tests hang locally, but run on GH action. I could not yet narrow down why this happens.
I ran a test on AWS and created a multi source asset (filter loss data by density) and it worked.

http://gfw-data-api-elb-feature-multiso-1131499710.us-east-1.elb.amazonaws.com/asset/7384576c-3324-4964-97ef-d4796120f61f/creation_options


app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -180,6 +180,7 @@ async def _merge_intensity_and_date_conf(
         source_type=RasterSourceType.raster,
         source_driver=RasterDrivers.geotiff,
         source_uri=[date_conf_uri, intensity_uri],
+        calc="[((A -((A>=30000) * 10000) - ((A>=20000) * 20000)) * (A>=20000)/255).astype('uint8'), ((A -((A>=30000) * 10000) - ((A>=20000) * 20000)) * (A>=20000) % 255).astype('uint8'), (((A>=30000) + 1)*100 + B).astype('uint8')]",  # this will not be used in PixETL but pydantic requires some input value
```
dmannarino, line 183: Why does Pydantic require something other than None? Isn't calc Optional?
In any case, if it is ignored I think it might be less confusing if we put in something more clearly place-holder like "This calc line does nothing" or something. Despite the comment, it's at the end of a long line and might not be seen immediately.

thomas-maschler: With my latest changes I enforce `calc` to be not optional when there are more than two input source uri, b/c normally pixetl would merge the two layers into one.

In this case, we don't use the creation options with pixetl but with the custom `build_rgb` script. I added the formular b/c this would be the numpy function we would have to use if we would use pixetl for this. But this would mean that pixetl could create output files with more than one band. and we are not quite there yet.

I am fine to put some other string here and leave it as a comment, just in case we need the formula later.

thomas-maschler, line 183:
```suggestion
        # [((A -((A>=30000) * 10000) - ((A>=20000) * 20000)) * (A>=20000)/255).astype('uint8'), ((A -((A>=30000) * 10000) - ((A>=20000) * 20000)) * (A>=20000) % 255).astype('uint8'), (((A>=30000) + 1)*100 + B).astype('uint8')]
        calc="This is a placeholder",  # this will not be used in PixETL but pydantic requires some input value
```

dmannarino (approved):

PR #153, opened 2021-02-27
Feature/multiband output

thomas-maschler:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

- support multi band output using pixetl v1.5.0
- Implement year_intensity symbology for TCL tile cache

- [ ] Yes
- [x] No


app/routes/__init__.py
```diff
@@ -11,6 +11,7 @@
 
 DATASET_REGEX = r"^[a-z][a-z0-9_-]{2,}$"
 VERSION_REGEX = r"^v\d{1,8}(\.\d{1,3}){0,2}?$|^latest$"
+DATE_REGEX = r"^\d{4}(\-(0?[1-9]|1[012])\-(0?[1-9]|[12][0-9]|3[01]))?$"
```
thomas-maschler, line 14: This is a broader version of the date regex which accepts (`yyyy`) year and full dates (`yyyy-mm-dd`). In most places only the full dates version was used. 
Let me think, if you see this as an issue, otherwise we might want to create entries here, once strict one and the one above accepting multiple inputs

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -126,6 +126,13 @@ async def raster_tile_cache_asset(
     assert source_asset_co.symbology is not None
     symbology_function = symbology_constructor[source_asset_co.symbology.type]
 
+    # We want to make sure that the final RGB asset is called after the implementation of the tile cache
+    if symbology_function == no_symbology:
+        source_asset_co.pixel_meaning = implementation
```
thomas-maschler, line 131: Following this logic through out the background task now. 
The final asset used to generate the tile cache will have same name as tile cache itself.
TODO: I might need to add some additional logic to catch cases where the tile cache implementation is called after an intermediate step, which would cause data overwrite and possible data corruption!

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -103,8 +104,73 @@ async def date_conf_intensity_symbology(
...
+    merge_function: Callable,
+) -> Tuple[List[Job], str]:
+    """Create Raster Tile Set asset which combines year raster and intensity
+    raster into one.
+
+    At native resolution (max_zoom) it will create intensity raster
+    based on given source. For lower zoom levels if will resample higher
+    zoom level tiles using mean resampling method. Once intensity raster
+    tile set is created it will combine it with source (year) raster
+    into rbg encoded raster.
```
thomas-maschler, line 172: update comment

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -158,6 +225,7 @@ async def date_conf_intensity_symbology(
 async def _merge_intensity_and_date_conf(
```
thomas-maschler, line 225: Technically, we could use PixETL for this one now as well and get rid of the C++ script. 
We should make this a seperate PR and time the differnce between the C++ and the PixETL implementation and go from there.

terraform/variables.tf
```diff
@@ -60,4 +60,9 @@ variable "git_sha" {
 }
 variable "lambda_analysis_workspace" {
   type = string
+}
+
+variable "data_lake_max_vcpus" {
+  type = number
+  default = 576
```
thomas-maschler, line 67: Increasing the number of CPUs of the compute environment to be able to run 6*98 cores at once.
Batch requests for SPOT CPUs with compete with EMR requests, so we shouldn't go much higher to assure our geotrellis jobs will always run.

app/models/pydantic/creation_options.py
```diff
@@ -108,6 +110,15 @@ class RasterTileSetAssetCreationOptions(StrictBaseModel):
...
     auxiliary_assets: Optional[List[UUID]] = None
+    photometric: Optional[PhotometricType] = None
+
+    @validator("no_data")
+    def validate_no_data(cls, v, values, **kwargs):
+        if isinstance(v, list):
+            assert len(v) != values.get(
+                "count"
+            ), "Length of no data list must much band count."
+        return v
```
dmannarino, line 121: You haven't changed the type of no_data to allow it to be a list yet.

app/routes/__init__.py
dmannarino: Doesn't strike me as an issue. I think this more flexible one is fine.

dmannarino (approved):

PR #154, opened 2021-03-02
Apply symbology in a separate minimal Batch job (outside of pixetl)

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Symbology is applied by running pixetl, though all IT really does is run gdaldem. This (at least sometimes) results in gdaldem running out of memory

Issue Number: GTC-1077

## What is the new behavior?

- gdaldem is run in its own Batch job, hopefully resolving the memory issues
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/app/tasks/raster_tile_cache_assets/symbology.py
+++ b/app/tasks/raster_tile_cache_assets/symbology.py
@@ -18,12 +18,11 @@ from app.models.pydantic.symbology import Symbology
 from app.settings.globals import MAX_CORES, MAX_MEM, PIXETL_DEFAULT_RESAMPLING
 from app.tasks import callback_constructor
 from app.tasks.raster_tile_cache_assets.utils import (
-    create_wm_tile_set_job,
     get_zoom_source_uri,
     reproject_to_web_mercator,
     tile_uri_to_tiles_geojson,
 )
-from app.tasks.raster_tile_set_assets.utils import JOB_ENV
+from app.tasks.raster_tile_set_assets.utils import JOB_ENV, create_gdaldem_job
 from app.tasks.utils import sanitize_batch_job_name
 from app.utils.path import get_asset_uri, split_s3_path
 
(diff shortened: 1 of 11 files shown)


app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -73,14 +112,28 @@ async def pixetl_symbology(
...
+            dataset,
+            version,
+            AssetType.raster_tile_set,
+            creation_options.dict(by_alias=True),
+            "epsg:3857",
+        )
+    )
+
+    pixel_meaning = f"{source_asset_co.pixel_meaning}_{source_asset_co.symbology.type}"
+    creation_options.pixel_meaning = pixel_meaning
```
thomas-maschler, line 130: move above line 108 and update creation_options only once.
You should set `source_uri` attribute to `source_asset_uri`. This way the correct value will be saved in the creation options and you don't need to pass around an extra parameter.

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -48,6 +47,46 @@ async def no_symbology(
         raise RuntimeError("No source URI set.")
 
 
+async def create_colormap_symbology_job(
```
thomas-maschler, line 50: why pull this code block into a seperate function? Is it a reused else where?
To me it looks like a continuation of the `pixetl_symbology` function. The actual job only gets generated in the gdaldemjob function, right?

thomas-maschler (changes requested):

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -73,14 +112,28 @@ async def pixetl_symbology(
...
+        get_asset_uri(
+            dataset,
+            version,
+            AssetType.raster_tile_set,
+            creation_options.dict(by_alias=True),
+            "epsg:3857",
+        )
+    )
+
+    pixel_meaning = f"{source_asset_co.pixel_meaning}_{source_asset_co.symbology.type}"
```
thomas-maschler, line 129: this will be the final tile set before running through the tiler. In my latest PR I give those tiles the same name as tile cache implementation. 

No need for action here. I can fix that when merging into my.

thomas-maschler (approved):

PR #162, opened 2021-03-19
Update raster query endpoint

jterry64:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

Raster analysis runs on separate endpoint 

Issue Number: N/A

## What is the new behavior?

- Raster analysis now runs on query endpoint, if the dataset default asset is a raster tile cache. 
- Uses new raster analysis with SQL interface.
- Now supports CSV download for raster analysis.

- [ ] Yes
- [X] No

### DIFF
--- a/app/models/enum/queries.py
+++ b/app/models/enum/queries.py
@@ -4,3 +4,7 @@ from enum import Enum
 class QueryFormat(str, Enum):
     json = "json"
     csv = "csv"
+
+class CsvDelimiter(str, Enum):
+    comma = ","
+    tab = "\t"
--- a/app/models/enum/queries.py
+++ b/app/models/enum/queries.py
@@ -4,7 +4,3 @@ from enum import Enum
 class QueryFormat(str, Enum):
     json = "json"
     csv = "csv"
-
-class CsvDelimiter(str, Enum):
-    comma = ","
-    tab = "\t"
--- a/app/routes/analysis/analysis.py
+++ b/app/routes/analysis/analysis.py
@@ -12,17 +12,12 @@ from fastapi.logger import logger
 from fastapi.responses import ORJSONResponse
 from httpx_auth import AWS4Auth
 
+from ..datasets.queries import _query_raster, _query_raster_lambda
 from ...models.enum.analysis import RasterLayer
 from ...models.enum.geostore import GeostoreOrigin
 from ...models.pydantic.analysis import ZonalAnalysisRequestIn, RasterQueryRequestIn
 from ...models.pydantic.geostore import Geometry
 from ...models.pydantic.responses import Response
-from ...responses import CSVStreamingResponse
-from ...settings.globals import (
-    AWS_REGION,
-    LAMBDA_ENTRYPOINT_URL,
-    RASTER_ANALYSIS_LAMBDA_NAME,
-)
 from ...utils.geostore import get_geostore_geometry
 from .. import DATE_REGEX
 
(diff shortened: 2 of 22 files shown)


dmannarino (approved):

app/routes/datasets/queries.py
```diff
@@ -302,3 +351,54 @@ async def _add_geometry_filter(parsed_sql, geometry: Geometry):
...
+    delimiter: Optional[str] = ",",
+) -> List:
+    # use default data type to get default raster layer for dataset
+    default_type = asset.creation_options["pixel_meaning"]
+    default_layer = (
+        f"{default_type}__{dataset}"
+        if default_type == "is"
+        else f"{dataset}__{default_type}"
+    )
+    sql = sql.lower().replace(f"from {dataset}", f"from {default_layer}")
```
thomas-maschler, line 371: technically, users should be able to add any table name, not just {dataset}

thomas-maschler: i am doing something similar in line 179 for the postgres queries

app/routes/datasets/queries.py
```diff
@@ -302,3 +351,54 @@ async def _add_geometry_filter(parsed_sql, geometry: Geometry):
...
+    return await _query_raster_lambda(geometry, sql, format, delimiter)
+
+
+async def _query_raster_lambda(
+    geometry: Geometry,
+    sql: str,
+    format: Optional[str] = "json",
+    delimiter: Optional[str] = ",",
```
thomas-maschler, line 379: Enum

app/routes/analysis/analysis.py
```diff
@@ -87,61 +76,42 @@ async def zonal_statistics_post(request: ZonalAnalysisRequestIn):
...
+
+    if end_date:
+        where_clauses.append(_get_date_filter(end_date, "<"))
+
+    where = " and ".join(filters)
+
+    query = f"select {selectors} from {base}"
+    if where:
+        query += f" where {where}"
+    query += f"group by {groups}"
```
thomas-maschler, line 108: OK, but I would prefer to stay consistent though out the code base and use SQL Alchemy core to compose SQL queries.

Here some examples in use:

https://github.com/wri/gfw-data-api/blob/master/app/crud/geostore.py#L32
https://github.com/wri/gfw-data-api/blob/master/batch/python/export_1x1_grid.py#L363-L375

thomas-maschler (changes requested):
Please also mark the analysis routes as deprecated:

https://fastapi.tiangolo.com/tutorial/path-operation-configuration/#deprecate-a-path-operation

app/routes/analysis/analysis.py
jterry64: @thomas-maschler ok I can do that, is the end result a SQL string or some object specific to sqlachemy?

app/models/enum/queries.py
```diff
@@ -4,3 +4,7 @@
 class QueryFormat(str, Enum):
     json = "json"
     csv = "csv"
+
+class CsvDelimiter(str, Enum):
+    comma = ","
+    tab = "\t"
```
thomas-maschler, line 10: Just realizing that there is already such an Enum

https://github.com/wri/gfw-data-api/blob/master/app/models/enum/creation_options.py#L26-L31

app/routes/analysis/analysis.py
```diff
@@ -76,42 +80,42 @@ async def zonal_statistics_post(request: ZonalAnalysisRequestIn):
...
     start_date: Optional[str],
     end_date: Optional[str],
 ):
-    base = sum_layers[0].value
-    selectors = ",".join([f"sum({lyr.value})" for lyr in sum_layers])
-    groups = ",".join([lyr.value for lyr in group_by])
+    if filters:
+        base = table(filters[0].value)
+    elif group_by:
+        base = table(group_by[0].value)
```
thomas-maschler, line 93: Do you need `.value` here. Since we mixed in `str` to the `Enum` you should be able to use the Enum straight away, no?

app/routes/datasets/queries.py
```diff
@@ -198,20 +200,20 @@ async def _query_table(
...
-def _orm_to_csv(data: List[RowProxy], delimiter=",") -> StringIO:
+def _orm_to_csv(data: List[RowProxy], delimiter: CsvDelimiter = CsvDelimiter.comma) -> StringIO:
     """Create a new csv file that represents generated data.
 
     Response will return a temporary redirect to download URL.
     """
     csv_file = StringIO()
 
-    wr = csv.writer(csv_file, quoting=csv.QUOTE_NONNUMERIC, delimiter=delimiter)
+    wr = csv.writer(csv_file, quoting=csv.QUOTE_NONNUMERIC, delimiter=delimiter.value)
```
thomas-maschler, line 216: same here, do you really need `.value`?

terraform/vars/terraform-dev.tfvars
```diff
@@ -4,4 +4,4 @@ service_url               = ""
 desired_count             = 1
 auto_scaling_min_capacity = 1
 auto_scaling_max_capacity = 5
-lambda_analysis_workspace = "feature-FanOutFanIn"
\ No newline at end of file
+lambda_analysis_workspace = "features-lat_lon"
```
thomas-maschler, line 7: No need to act right now, but we should look into otherwise to link to the right lambda function in dev

- integrate lambda into api repo?
- always push to production and build against distinct deployment versions?
- other ideas?

thomas-maschler (approved):

app/routes/analysis/analysis.py
thomas-maschler: the string representation will be a SQL expression ie : `str(query)`

PR #170, opened 2021-04-19
Direct ORJSON Serialization for Query Endpoint

jterry64:
This is something I discovered when I initially did performance testing for postgres migration. Using the default response serialization makes FastAPI call its custom jsonencoder, which runs very slow because it basically overkills checking every possible type conversion. 

For query, where we have relatively simple return types, but can return massive lists, we should just supplement the basic ORJSON conversion with a few known types. This runs way way faster. 

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Using default JSON serialization from FastAPI for query endpoint. From performance testing, this can be very slow for queries that return a large number of results, since for each row it does extensive checks for possible types.

## What is the new behavior?
Rely just on default serialization directly from ORJSON, which is way fast. Supplement default serialization with just bare minimum of known types. This should be fairly straightforward for query, since it'll only return basic database types. 

- [ ] Yes
- [X] No


app/routes/datasets/queries.py
```diff
@@ -95,9 +98,10 @@ async def query_dataset(
     else:
         geometry = None
 
-    data: List[RowProxy] = await _query_dataset(dataset, version, sql, geometry)
+    data: List[Dict[str, Any]]= await _query_dataset(dataset, version, sql, geometry)
+    serialized_data = orjson.dumps(data, default=jsonencoder_lite())
 
-    return Response(data=data)
+    return Response(content=serialized_data, media_type="application/json")
```
thomas-maschler, line 104: Which `Response` class do you mean to reference here?
Right now you have `from ...models.pydantic.responses import Response` in scop. This is one of our pydantic models which assures that our responses follow the JSEND pattern.

It seems like you want to return a standard FastAPI `Response` which is not in scope. If doing so, please make sure you place the query results under the `data` key and add another root level key `"status":"success"`

thomas-maschler (changes requested):

dmannarino (approved):

PR #171, opened 2021-04-21
Feature/api key

thomas-maschler:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Introduce API Keys
- start implementing test suite v2
-

- [x] Yes
- [ ] No

query endpoint requires API token


dmannarino (approved):

app/crud/api_keys.py
```diff
@@ -0,0 +1,65 @@
...
+    user_id: str,
+    alias: Optional[str],
+    organization: str,
+    email: str,
+    domains: List[str],
+) -> ORMApiKey:
+
+    # If a simple string is used for domains, sqlalchemy will still inject this into db,
+    # using every character as a list item. We don't want this to happen
+    assert isinstance(domains, list), "Domains must be of type List[str]"
```
jterry64, line 19: shouldn't this already be caught by the List[str] type for the parameter? or are there paths that a string will come instead of a list?

app/routes/authentication.py
```diff
@@ -0,0 +1,129 @@
...
+
+    else:
+        return {
+            "access_token": token,
+            "token_type": "bearer",
+        }
+
+
+@router.post("/apikey", tags=["Authentication"], status_code=201)
+async def create_apikey(
```
jterry64, line 49: nitpick: should this be 'api_key' instead of a 'apikey'?

app/routes/authentication.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    input_data = request.dict(exclude_none=True, by_alias=True)
+
+    row: ORMApiKey = await api_keys.create_api_key(user_id=user_id, **input_data)
+
+    return ApiKeyResponse(data=row)
+
+
+@router.get("/apikeys", tags=["Authentication"])
```
jterry64, line 65: same as above, should it be 'api_keys' instead of 'apikeys'?

jterry64 (approved):
Looks good, just a few minor comments. I'm assuming in the future we'll merge tests and tests_v2?

app/crud/api_keys.py
thomas-maschler: I noticed while testing that when adding strings it will become funky. `List[str]` is only a type hint and not enforced. I prefer to catch this for sure before injecting it into the DB. You never know.

app/routes/authentication.py
thomas-maschler: i am back and forth on this, too. I went for the merged version in the path to keep it simple

app/routes/authentication.py
thomas-maschler: test_v2 will replace test at one point. This will be a rewrite of all tests to improve test coverage and reduce test runtime. F

PR #172, opened 2021-04-22
improve tagging of AWS resources

thomas-maschler:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

- Improve AWS resource tagging for better cost monitoring

- [ ] Yes
- [X] No

### DIFF
--- a/app/models/pydantic/jobs.py
+++ b/app/models/pydantic/jobs.py
@@ -21,6 +21,7 @@ from .base import StrictBaseModel
 
 
 class Job(StrictBaseModel):
+    dataset: str  # used for tagging resources
     job_name: str
     job_queue: str
     job_definition: str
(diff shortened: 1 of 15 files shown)


jterry64 (changes requested):
Just checked out how grouping by tag works in cost explorer, and it only lets you group by tag name, with no special querying. This might be pretty annoying for geotrellis tables, since we make so many for each analysis. E.g. for GADM GLAD, we'd have to add up the cost of 10 different tags to get the full amount. 

Since the data API is unaware of geotrellis, maybe we could make use of the "tags" in the dataset metadata to have the client be able to add tags to jobs? Not sure what that's used for right now.

thomas-maschler:
The tags in the metadata were more meant for data catalogs to describing the dataset basic nature of the dataset.

However, I don't think that we will actually get a deeper understanding on how Batch resources are used by dataset out of this anyways.

Tags for Ec2 instances used by Batch are assigned at the compute environment level. They are used by different dataset and are generally unaware of the kind of job they are executing. 

We can add tags for each batch jobs, but they won't expand down to the EC2 instance. So from a cost monitoring perspective these tags will probably have not much of an impact. 

But what we should be able to see if which EC2 costs are caused by Batch and which by EMR.

jterry64 (approved):
Ok that makes sense, approved

PR #177, opened 2021-05-03
Fix Cloudfront CNAME for dev environment

jterry64:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Multiple feature branches on dev can't deploy the cloudfront cache via terraform, since the CNAME is hardcoded so will be a duplicate.

Issue Number: N/A

## What is the new behavior?
Include the feature branch name in the CNAME to make it unique per feature branch.

- [ ] Yes
- [X] No

### DIFF
--- a/terraform/cloudfront.tf
+++ b/terraform/cloudfront.tf
@@ -2,7 +2,7 @@ resource "aws_cloudfront_distribution" "data_api" {
   enabled             = true
   is_ipv6_enabled     = true
   price_class         = "PriceClass_All"
-  aliases             = [replace(var.service_url, "https://", "")]
+  aliases             = [var.environment == "dev"? "${local.project}-data-api.globalforestwatch.org" : replace(var.service_url, "https://", "")]
 
   origin {
     domain_name = module.fargate_autoscaling.lb_dns_name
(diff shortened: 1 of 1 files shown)


dmannarino (approved):

terraform/cloudfront.tf
```diff
@@ -2,7 +2,7 @@ resource "aws_cloudfront_distribution" "data_api" {
   enabled             = true
   is_ipv6_enabled     = true
   price_class         = "PriceClass_All"
-  aliases             = [replace(var.service_url, "https://", "")]
+  aliases             = [var.environment == "dev"? "${local.project}-data-api.globalforestwatch.org" : replace(var.service_url, "https://", "")]
```
thomas-maschler, line 5: Hmm, I don;t think this will work. We don't point any `globalforestwatch.org` domain to api services in the dev account. Can you leave the aliases blank?
```suggestion
  aliases             = [var.environment == "dev" ? null : replace(var.service_url, "https://", "")]
```

thomas-maschler (changes requested):

terraform/cloudfront.tf
jterry64: You can't leave it blank unfortunately, terraform will throw an error, so that's why I just added a dummy CNAME. You can't actually go the URL, you just have to go the CloudFront distribution URI. But it's ok with having an unreachable CNAME.

thomas-maschler: terraform docs say it is optional, so null should normally work. Odd. Ok, what ever works, but please leave a comment in the code on why you picked this name for later reference

PR #179, opened 2021-05-05
Zonal Analysis Transformer Fixes

jterry64:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Broken in some cases because of refactoring.

## What is the new behavior?
Properly uses new interface.

- [ ] Yes
- [X] No

### DIFF
--- a/app/routes/datasets/queries.py
+++ b/app/routes/datasets/queries.py
@@ -2,7 +2,6 @@
 import csv
 import re
 from io import StringIO
-from json import JSONDecodeError
 from typing import Any, Dict, List, Optional, Tuple, cast
 from urllib.parse import unquote
 from uuid import UUID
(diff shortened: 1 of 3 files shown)


app/routes/datasets/queries.py
```diff
@@ -400,9 +399,8 @@ async def _query_raster_lambda(
     except httpx.TimeoutException:
         raise HTTPException(500, "Query took too long to process.")
 
-    try:
-        response_data = response.json()["body"]
-    except (JSONDecodeError, KeyError):
+    response_data = response.json()["body"]
+    if response_data["status"] != "success":
```
thomas-maschler, line 403: I would use `.get` here in case the response has something unexpected. Will Lambda functions always have a body object or can it also be something different? If yes, you might want to catch this as well.

```suggestion
    response_data = response.json()["body"]
    if response_data.get("status") != "success":
```

thomas-maschler (changes requested):

app/routes/datasets/queries.py
jterry64: @thomas-maschler I think it'll always have a body if the lambda returns successfully, but you're right that I'm not catching if the lambda returns a general error (i.e. couldn't run the lambda for whatever reason). So I added a check of the status code from the response, and just generally check the body has a status and a data field otherwise it fails.

dmannarino (approved):

thomas-maschler (approved):

PR #195, opened 2021-05-20
Allow returning multiple formats from query

jterry64:
This is a feature request from Forest Watcher, who would like to be able to also get CSV from the query endpoint without it being a file download. I made just as parity with RW API, where you can submit a parameter for the format type (default is JSON). 

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Only way to get CSV is via download endpoint.

## What is the new behavior?

- Allow query parameter to get CSV response from query endpoint
- Some small bug fixes I found for OTF queries

- [ ] Yes
- [X] No

### DIFF
--- a/app/models/pydantic/query.py
+++ b/app/models/pydantic/query.py
@@ -1,5 +1,7 @@
 from typing import Optional
 
+from app.models.enum.creation_options import Delimiters
+from app.models.enum.queries import QueryFormat
 from app.models.pydantic.base import StrictBaseModel
 from app.models.pydantic.geostore import Geometry
 
@@ -7,3 +9,5 @@ from app.models.pydantic.geostore import Geometry
 class QueryRequestIn(StrictBaseModel):
     geometry: Optional[Geometry]
     sql: str
+    format: QueryFormat = QueryFormat.json
+    delimter: Delimiters = Delimiters.comma
--- a/app/models/pydantic/query.py
+++ b/app/models/pydantic/query.py
@@ -1,7 +1,6 @@
 from typing import Optional
 
 from app.models.enum.creation_options import Delimiters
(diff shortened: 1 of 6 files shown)


app/models/pydantic/query.py
```diff
@@ -1,9 +1,13 @@
...
+from app.models.enum.queries import QueryFormat
 from app.models.pydantic.base import StrictBaseModel
 from app.models.pydantic.geostore import Geometry
 
 
 class QueryRequestIn(StrictBaseModel):
     geometry: Optional[Geometry]
     sql: str
+    format: QueryFormat = QueryFormat.json
+    delimter: Delimiters = Delimiters.comma
```
thomas-maschler, line 13: The payload should now be different.
Format is part of the path and does not be part of the payload anymore.
Delimiter is only required for CSV downloads and should live in a subclass

app/routes/datasets/queries.py
```diff
@@ -115,19 +150,65 @@ async def query_dataset(
...
-    return Response(data=data)
+    csv_data: StringIO = await _query_dataset_csv(
+        dataset, version, sql, geometry, delimiter=delimiter
+    )
+    return CSVStreamingResponse(iter([csv_data.getvalue()]), download=False)
 
 
 @router.post(
-    "/{dataset}/{version}/query",
+    "/{dataset}/{version}/query/json",
```
thomas-maschler, line 206: There also needs to be a redirect for this URL, right?

thomas-maschler (changes requested):

dmannarino (approved):

jterry64:
@thomas-maschler fixed your comments

PR #202, opened 2021-05-25
Allow CSV vector sources

jterry64:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

CSV imports for vectors will just fail, and for vector imports only one source URI is allowed.

## What is the new behavior?

Allow vector CSV import for many source URIs by following the same method as table imports, but using ogr2ogr to upload to postgres.

- [ ] Yes
- [X] No

## Other information

If we want to add partitions/clustering to vector source assets now too, I will add them as a separate PR, since it'd be a big refactor and I want to just get the data uploaded in staging so we can start prototyping with it.

### DIFF
--- a/app/tasks/vector_source_assets.py
+++ b/app/tasks/vector_source_assets.py
@@ -92,11 +92,6 @@ async def vector_source_asset(
 
         load_data_parents = [job.job_name for job in load_vector_data_jobs]
     else:
-        if len(source_uris) != 1:
-            raise AssertionError(
-                "Non-CSV vector sources require one and only one input file"
-            )
-
         # AWS Batch jobs can't have more than 20 parents. In case of excessive
         # numbers of layers, create multiple "queues" of dependent jobs, with
         # the next phase being dependent on the last job of each queue.
--- a/app/tasks/vector_source_assets.py
+++ b/app/tasks/vector_source_assets.py
@@ -9,7 +9,7 @@ from ..models.pydantic.creation_options import VectorSourceCreationOptions
 from ..models.pydantic.jobs import GdalPythonImportJob, Job, PostgresqlClientJob
 from ..utils.path import get_layer_name, is_zipped
 from . import Callback, callback_constructor, writer_secrets
-from .batch import execute, BATCH_DEPENDENCY_LIMIT
+from .batch import BATCH_DEPENDENCY_LIMIT, execute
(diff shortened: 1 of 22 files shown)


app/tasks/vector_source_assets.py
```diff
@@ -60,48 +59,88 @@ async def vector_source_asset(
...
+                callback=callback,
+            )
+            load_vector_data_jobs.append(job)
+
+        load_data_parents = [job.job_name for job in load_vector_data_jobs]
+    else:
+        if len(source_uris) != 1:
+            raise AssertionError(
+                "Non-CSV vector sources require one and only one input file"
+            )
```
thomas-maschler, line 98: It would actually be best to check this as early as possible to give users immediate feedback.
You could make this a `validator` in the VectorSourceCreationOptions 


```python
@validator("source_uri")
        if values.get("source_driver") != VectorDrivers.csv:
            assert len(v), "Non-CSV vector sources require one and only one input file"
        return v
```

batch/scripts/create_vector_schema.sh
```diff
@@ -29,7 +29,8 @@ ogr2ogr -f "PostgreSQL" PG:"password=$PGPASSWORD host=$PGHOST port=$PGPORT dbnam
      "$LOCAL_FILE" "$SRC_LAYER" \
      -nlt PROMOTE_TO_MULTI -nln "$DATASET.$VERSION" \
      -lco GEOMETRY_NAME="$GEOMETRY_NAME" -lco SPATIAL_INDEX=NONE -lco FID="$FID_NAME" \
-     -t_srs EPSG:4326 -limit 0
+     -oo GEOM_POSSIBLE_NAMES="$GEOMETRY_NAME" -oo KEEP_GEOM_COLUMNS=NO \
+     -t_srs EPSG:4326 -s_srs EPSG:4326 -limit 0
```
thomas-maschler, line 33: This makes sense, but please leave a comment on why you set `s_srs` since it is not right away clear for an outsider to understand the logic.

thomas-maschler (changes requested):

dmannarino (approved):

jterry64:
@thomas-maschler fixed your comments

thomas-maschler (approved):

PR #205, opened 2021-05-28
Add timeout parameter for table source asset import

jterry64:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Timeout for table asset import is always 2 hours for each batch job.

Issue Number: N/A

## What is the new behavior?
Add a parameter to increase the timeout for big imports.

- [ ] Yes
- [X] No

### DIFF
--- a/app/models/pydantic/creation_options.py
+++ b/app/models/pydantic/creation_options.py
@@ -5,7 +5,7 @@ from uuid import UUID
 from pydantic import Field, root_validator, validator
 from pydantic.types import PositiveInt, StrictInt
 
-from ...settings.globals import PIXETL_DEFAULT_RESAMPLING
+from ...settings.globals import DEFAULT_JOB_DURATION, PIXETL_DEFAULT_RESAMPLING
 from ..enum.assets import AssetType, is_default_asset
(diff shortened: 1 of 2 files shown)


app/models/pydantic/creation_options.py
```diff
@@ -228,6 +228,7 @@ class TableAssetCreationOptions(StrictBaseModel):
         "when geographic columns are present. "
         "Disable this option by setting value to `false`",
     )
+    timeout: int = DEFAULT_JOB_DURATION
```
thomas-maschler, line 231: Since you add a default values here this should work, but please double check that we don't get an error when retrieving an old creation option where the value was not yet set.

thomas-maschler (approved):
Looks good. We should add the option to all creation options to overwrite the default value. We can make this a seperate ticket

dmannarino (dismissed):

app/models/pydantic/creation_options.py
jterry64: Looks like it works, just adds the default value

PR #223, opened 2021-07-22
Retry Batch jobs on exit code 137

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Batch jobs retry strategy will only catch OOM situations when it so happened that the primary process was killed

Issue Number: N/A

## What is the new behavior?

- In tandem with changes to pixetl, retry on Batch jobs exiting with code 137 (process was killed) instead of checking the container's reason
-
-

- [ ] Yes
- [x] No


thomas-maschler (approved):
Looks fine to me. Is it really necessary to processify all the python scripts in the batch folder?
I thought that as long as the main process gets kills, the contain always exists with 137. The batch scripts use a different multiprocessing approach than what we use in pixETL.

PR #226, opened 2021-08-06
Integrated layer support

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Missing support for various things needed by the integrated alerts dataset (16-bit per channel raster tile caches, multi-band inputs, 

Issue Number: N/A

## What is the new behavior?

- Has support for 8-bit and 16-bit integrated alert symbologies
- Allows one to direct pixetl to union the extent of bands instead of taking their intersection
- Disables validation preventing processing of multi-band input rasters
- Generates a generic calc string when creating a raster tile cache from multi-band tile sets
- Fixes a bug where adjust_num_processes will fail if a Batch job's exitCode is not populated (as is the case when a spot instance has been taken away)
- Add hacked gdal2tiles copies for 8- and 16-bit tile caches with 4 channels
- Update pixetl, mostly to enable taking multi-band inputs and unioning their extents

- [ ] Yes
- [x] No


app/models/pydantic/creation_options.py
```diff
@@ -127,10 +128,11 @@ def validate_no_data(cls, v, values, **kwargs):
 
     @validator("band_count")
     def validate_band_count(cls, v, values, **kwargs):
+        # Should we really require this or make pixetl more flexible?
```
jterry64, line 131: you mean like allow resampling or other non-calc operations across multiple bands?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -62,6 +64,15 @@ async def raster_tile_cache_asset(
         ).replace("{tile_id}.tif", "tiles.geojson")
     ]
 
+    # Create a generic calc string when a raster tile cache is being
+    # created from a multi-band raster tile set.
+    band_count = source_asset.creation_options["band_count"]
+    if band_count > 1:
+        bands_string = str(list(string.ascii_uppercase[:band_count])).replace("'", "")
```
jterry64, line 71: Maybe put a comment describing what you're doing here for the bands_string, since there's a lot going on on this line

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -62,6 +64,15 @@ async def raster_tile_cache_asset(
...
+        calc: Optional[str] = f"np.ma.array({bands_string})"
```
jterry64, line 72: instead of requiring this np.ma.array always in the calc string, should we just automatically wrap the calc string in np.ma.array?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -71,16 +82,18 @@ async def raster_tile_cache_asset(
...
                 "compute_stats": False,
                 "compute_histogram": False,
                 "symbology": Symbology(**symbology),
+                "subset": None,
             },
         }
     )
 
     # If float data type, convert to int in derivative assets for performance
+    # FIXME: Make this work for multi-band inputs
```
jterry64, line 96: did you mean to do this now or later?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -147,6 +161,18 @@ async def raster_tile_cache_asset(
...
         job_list += symbology_jobs
 
+        # FIXME: Surely this could be done better.
+        # Maybe make a lookup table of bit depths for symbology types?
+        # The larger question is should we infer the bit depth from the
+        # symbology (or even from the data in the source) or give people
+        # the flexibility/responsibility of specifying it?
+        # Doing this for the moment as it's the only case in which we
+        # want 16-bit PNGs
+        if symbology_function == date_conf_intensity_multi_16_symbology:
```
jterry64, line 171: Agreed, why don't you just create a lookup around where you define the symbology constructor lookup?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -180,7 +207,7 @@ async def raster_tile_cache_validator(
             detail="Dataset and version of source asset must match dataset and version of current asset.",
         )
 
-    if source_asset.creation_options.get("band_count", 1) > 1:
-        raise HTTPException(
-            status_code=400, detail="Cannot apply colormap on multi-band image."
-        )
+    # if source_asset.creation_options.get("band_count", 1) > 1:
```
jterry64, line 210: remove this if it's no longer necessary

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -118,6 +118,77 @@ async def colormap_symbology(
...
+        intensity_co,
+        zoom_level,
+        max_zoom,
+        jobs_dict,
+        # What we want is a value of 55 (max intensity) anywhere there is
+        # an alert in any system. We can't just do
+        # "((A > 0) | (B > 0) | (C > 0)) * 55" because "A | B" includes only
+        # those values unmasked in both A and B. So first replace masked
+        # values with 0 and then re-mask them later
+        "np.ma.array([((A.filled(0) >> 1) > 0) * 55, ((B.filled(0) >> 1) > 0) * 55, ((C.filled(0) >> 1) > 0) * 55])",
```
jterry64, line 153: would it be possible to write complex calcs like these as a multiline string with indenting, so it's easier to mentally parse what's going on?

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -118,6 +118,77 @@ async def colormap_symbology(
...
+    intensity_co = source_asset_co.copy(deep=True, update={"data_type": DataType.uint8})
+    return await _date_intensity_symbology(
+        dataset,
+        version,
+        pixel_meaning,
+        intensity_co,
+        zoom_level,
+        max_zoom,
+        jobs_dict,
+        "np.ma.array([((A.filled(0) >> 1) > 0) * 31, ((B.filled(0) >> 1) > 0) * 31, ((C.filled(0) >> 1) > 0) * 31])",
```
jterry64, line 186: Why 31 instead of 55 here? Add a comment maybe? Also, perhaps store these numbers as constants somewhere

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -252,6 +323,232 @@ async def _date_intensity_symbology(
...
+    #         for not detected.
+
+    # But we've got three systems, and we want the minimum (first detection)
+    # in any. np.minimum doesn't exclude masked values, so we have to replace
+    # them with something bigger than our data... and then re-mask the
+    # previously masked values afterwards. So unless I'm mistaken, just to
+    # get the minimum of the three alert systems requires something like this:
+    # Oye. What a monster...
+
+    FIRST_ALERT = "(np.ma.array((np.ma.array(np.minimum(A.filled(65535), np.minimum(B.filled(65535), C.filled(65535))), mask=(A.mask & B.mask & C.mask)).filled(0)), mask=(A.mask & B.mask & C.mask)))"
```
jterry64, line 381: definitely could use a multiline string here too

jterry64: why are all the vars in this function in constant uppercase format?

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -383,7 +680,7 @@ async def _merge_intensity_and_year(
         source_driver=RasterDrivers.geotiff,
         source_uri=[intensity_uri, year_uri],
         band_count=3,
-        no_data=0,
+        no_data=0,  # FIXME: Shouldn't this be [0, 0, 0]?
```
jterry64, line 683: does it work correctly?

batch/python/raster_tile_cache.py
```diff
@@ -89,7 +89,7 @@ def get_input_tiles(prefix: str) -> List[Tuple[str, str]]:
     return tiles
 
 
-def create_tiles(args: Tuple[Tuple[str, str], str, str, str, str, int, bool, int]):
+def create_tiles(args: Tuple[Tuple[str, str], str, str, str, str, int, bool, int, int]):
```
jterry64, line 92: I guess this was already like this, but why is the type here this ridiculously long tuple type lol? Maybe we should consider using named tuples, or defining this as an object.

jterry64 (changes requested):
@dmannarino left some comments, mostly just on style. Also, would it be possible to add some tests for this?

solomon-negusse (approved):

app/models/pydantic/creation_options.py
dmannarino: No, I mean that right now we don't require a calc string for a one-band raster and assume that the client wants an output band identical to the input band. But why not do the same for multiple-band inputs? I implemented this (generating a calc string that just passes any input bands through unmodified) in the data API code that creates a raster tile cache because I had to, but it seems to me a reasonable thing to do as a general default. I chickened-out of doing it across the board because it felt bad to eliminate a validator, but if we were to have pixetl pass bands through unmodified by default it wouldn't be validating anything useful. I think I'll leave this check as-is for now, and remove it when pixetl is capable of taking a multi-band input without a calc string.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Done

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Hmm. It's not ALWAYS required (the details are not worth going into), but I can't think of any harm to wrapping the calc string in an extra np.ma.array, and it would mean the user wouldn't have to remember to do so. Sounds good, though I think the change should be made in pixetl. When I've done that I'll modify any relevant calc strings in the data API code.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Later. Single-band float datasets will continue to work as before. Do we have any multi-band float datasets on the horizon @manukala6 ?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Done.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Replaced with a better test.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Done

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: I think I've improved it quite a bit.

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: Explanation added (it's because 31 is the largest value we can store in 5 bits, which is all we have for each channel when packed into the 16-bit alpha band).

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: Eh, good question. Fixed!

batch/python/raster_tile_cache.py
dmannarino: Ugh, the argument is required to be a tuple by the process execution method we're using. Will try making a NamedTuple.

app/models/pydantic/creation_options.py
jterry64: sounds good to me

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
jterry64: gotcha, we can do this in a separate PR then

jterry64 (approved):
Looks good, I think the rest of can be handled later, thanks for making the changes!

PR #231, opened 2021-08-31
Only allow assets in a single grid to be analyzed in an OTF query

jterry64:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

An OTF query can include raster assets in multiple grids, as long as it's upsampling. However, it turns out the dynamic resampling was off from the static resampling in pixetl. Moreover, now that many of our assets are resampled to multiple grids, a lot of the fields are being listed multiple times for each resolution.

## What is the new behavior?

- OTF will only analyze rasters strictly in the same grid, and this grid will be determined by the primary asset being analyzed
- The fields endpoint for rasters will now only list rasters available in the grid of the primary asset

- [X] Yes
- [ ] No

This potentially makes it so contextual layers not resampled to 10/100000 are no longer available for OTF. But no one should be using those yet, and anyway we should be resampling them if we need them. @manukala6 could you confirm if you think this might be an issue?

### DIFF
--- a/app/routes/datasets/queries.py
+++ b/app/routes/datasets/queries.py
@@ -573,20 +573,20 @@ async def _query_raster(
         if default_type == "is"
         else f"{dataset}__{default_type}"
     )
+    grid = asset.creation_options["grid"]
 
     sql = re.sub("from \w+", f"from {default_layer}", sql, flags=re.IGNORECASE)
-    return await _query_raster_lambda(geostore.geojson, sql, format, delimiter)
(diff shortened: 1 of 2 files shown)


solomon-negusse (approved):

dmannarino (approved):

manukala6 (approved):
Shouldn't be an issue. We created 10/100000 tile sets for all contextual layers for the Q3 update.

PR #251, opened 2021-10-28
Allow S3 and GS prefixes to be used as src_uri items

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- Specifying an S3 prefix/folder as a source_uri item (for example:s3://some_bucket/my_data/) is broken, and it looks like it was never implemented for GCS sources. This means we need to pre-generate a tiles.geojson for every data import.
- Pixetl breaks when the ratio of CPUs to tiles is large, such as when running on a subset of a dataset or when resuming a mostly-complete job after our EC2 is recalled

Issue Number: GTC-1331, GTC-1261

## What is the new behavior?

- One can use a gs:// or s3:// URI for src_uris!
- Pixetl will succeed even when the ratio of CPUs to tiles is large

- [ ] Yes
- [x] No

## Other information

Also updated quite a lot of dependencies between the data API and pixetl


app/utils/aws.py
```diff
@@ -67,6 +73,47 @@ async def head_s3(bucket: str, key: str) -> bool:
     return response.status_code == 200
 
 
+def get_aws_files(
```
jterry64, line 76: If this is blocking, and there's no default limit on the number of files, could someone past a prefix with a bazillion files and just completely block the service?

jterry64 (approved):
Looks good, just had one question above

app/utils/aws.py
dmannarino: Mmm, you're right. It's not the RIGHT solution, but for the moment I will add a default limit of 10. This means if you specify a prefix filled with 10 things other than valid files, there's a chance you will get a misleading error message. But hey, they should clean up their junk! Also will create a ticket to lay out everything wrong with this approach.

dmannarino: Follow-up ticket: https://gfw.atlassian.net/browse/GTC-1434

solomon-negusse (approved):

app/utils/aws.py
jterry64: Oh I missed where you set it! Sounds good.

PR #255, opened 2021-11-12
Revision query

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- adds revision filtering to queries

- [ ] Yes
- [x] No

### DIFF
--- a/app/routes/datasets/queries.py
+++ b/app/routes/datasets/queries.py
@@ -23,6 +23,7 @@ from sqlalchemy.sql import and_
 # from ...authentication.api_keys import get_api_key
 from ...application import db
 from ...crud import assets
+from ...errors import RecordNotFoundError
 from ...models.enum.assets import AssetType
 from ...models.enum.creation_options import Delimiters
 from ...models.enum.geostore import GeostoreOrigin
@@ -332,6 +333,7 @@ def _get_query_type(default_asset: AssetORM, geostore: Optional[GeostoreCommon])
     if default_asset.asset_type in [
         AssetType.geo_database_table,
         AssetType.database_table,
+        AssetType.revision,
     ]:
         return QueryType.table
     elif default_asset.asset_type == AssetType.raster_tile_set:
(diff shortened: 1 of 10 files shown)


app/routes/datasets/queries.py
```diff
@@ -552,6 +563,102 @@ async def _add_geometry_filter(parsed_sql, geometry: Geometry):
...
+        }
+    else:
+        parsed_sql[0]["RawStmt"]["stmt"]["SelectStmt"]["whereClause"] = {
+            "BoolExpr": {"boolop": 0, "args": revision_filters}
+        }
+
+    return parsed_sql
+
+
+async def _filter_by_revision_operation(revision_asset: AssetORM, filter_type: str):
```
jterry64, line 622: What do the output filters look like for this? It seems like it'd be something like

`gfw_version != 'v1' AND gfw_version !='v2' AND ...`

If so, I think it'd make more sense to just make an IN statement, and making it like

`gfw_version NOT IN ('v1', v2', ...)`

I think this runs faster in SQL when there's a lot of filters.

app/routes/datasets/queries.py
```diff
@@ -552,6 +563,102 @@ async def _add_geometry_filter(parsed_sql, geometry: Geometry):
...
+            revision_asset = None
+
+    sql_where = parsed_sql[0]["RawStmt"]["stmt"]["SelectStmt"].get("whereClause", None)
+
+    if revision_asset is None:
+        revision_filters = []
+    elif revision_asset.version == latest_revision:
+        revision_filters = await _filter_by_revision_operation(revision_asset, "latest")
+    else:
+        revision_filters = await _filter_by_revision_operation(
```
jterry64, line 603: I think you need to pass the latest revision history somehow, since the revision history for the intermediate revision won't include revisions after it

app/routes/datasets/queries.py
```diff
@@ -552,6 +563,102 @@ async def _add_geometry_filter(parsed_sql, geometry: Geometry):
...
+            revision["dataset"], revision["version"]
+        )
+        delete_revision = asset.creation_options.get("delete_version", None)
+
+        if delete_revision is not None:
+            delete_revisions.append(delete_revision)
+        else:
+            append_revisions.append(revision["version"])
+
+    if filter_type == "latest":
```
jterry64, line 636: I don't think you need this fork in logic based on a special keyword. I think both cases can latest and intermediate cases can apply the same logic. Pass both latest revision history and the current revision, then:

- Fill append_revisions with any append revisions that come after the current revision in the latest revision history (so this will be empty for the latest revision, but populated for intermediate revisions)
- Fill delete_revisions with all delete revisions
- Combine append_revisions and delete_revisions into a set to delete any duplicates
- Make your filter:

`WHERE gfw_version NOT IN ([revision for revision in combined_revisions])`

app/routes/datasets/versions.py
```diff
@@ -106,8 +108,11 @@ async def add_new_version(
 
     input_data["creation_options"] = creation_options
 
-    # Everything else happens in the background task asynchronously
-    background_tasks.add_task(create_default_asset, dataset, version, input_data, None)
+    if creation_options.get("delete_version", None):
```
jterry64, line 111: Is this to make an asset record for delete operations? Think I overlooked this, thanks!

jterry64 (changes requested):

app/routes/datasets/queries.py
solomon-negusse: That's neat, thanks!

jterry64 (approved):

PR #256, opened 2021-11-17
Feature/revisions

jterry64:
Opening to compare changes until we're ready


app/crud/versions.py
```diff
@@ -72,6 +73,13 @@ async def create_version(dataset: str, version: str, **data) -> ORMVersion:
     if data.get("is_latest"):
         await _reset_is_latest(dataset, version)
 
+    creation_options = data.get("creation_options")
+
+    # if revision, merge with previous
+    if creation_options and creation_options["source_type"] == SourceType.revision:
+        prev_version = await get_version(dataset, creation_options["revision_on"])
+        data["metadata"] = {**data["metadata"], **prev_version.metadata}
```
dmannarino, line 81: Should this be `data["metadata"] = {**prev_version.metadata, **data["metadata"]}` instead? It looks to me like the old metadata will take precedence over the new, but I could be mis-reading it.

app/models/pydantic/jobs.py
```diff
@@ -27,10 +27,10 @@ class Job(StrictBaseModel):
...
     command: List[str]
     environment: List[Dict[str, str]] = []
-    vcpus: int
-    memory: int
-    attempts: int
-    attempt_duration_seconds: int
+    vcpus: int = 8
+    memory: int = 3200
+    attempts: int = 1
+    attempt_duration_seconds: int = 7200
```
dmannarino, line 33: I think we specify these in all the child classes so it shouldn't matter, but curious why add defaults?

app/routes/datasets/queries.py
```diff
@@ -370,21 +370,31 @@ async def _query_table(
...
         parsed = await _add_geometry_filter(parsed, geometry)
 
+    # filtering on revisions
+    parsed = await _add_revision_filter(parsed, dataset, version)
+
     # convert back to text
     sql = RawStream()(Node(parsed))
 
     try:
+        logger.info(f"Executing query: {sql}")
```
dmannarino, line 397: Maybe make logger.debug?

PR #269, opened 2021-12-29
Run select raster import cases with GDAL CLI utilities instead of through pixetl

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Specifying a number of resampling methods (including "min", "max", and "mode") when creating a raster tile cache produces data with anomalies due to a rasterio bug
- Performance when importing rasters and not specifying a calc string can be improved by using the GDAL command line utilities instead of pixetl

Issue Number: https://gfw.atlassian.net/browse/GTC-1563

## What is the new behavior?

- Select raster import phases (specifically when resampling from zoom level to zoom level with no calc function) are run through a new resample.py script which uses the GDAL CLI tools. This improves performance and circumvents the rasterio bug(s) WRT other resampling methods

- [ ] Yes
- [x] No

## Other information

The number of processes used is tuned to use almost 100% memory when processing integrated alerts. A more comprehensive and robust method of saturating (but not exceeding) system resources regardless of dataset should be found in the future.


jterry64 (changes requested):
One general question: why did you add this in this repo instead of just allowing an option to directly call GDAL CLI in PixETL? I feel like now we've duplicated responsibilities here, and will have to maintain a whole separate way of processing tiles with GDAL.

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -313,14 +293,16 @@ async def date_conf_intensity_multi_8_symbology(
...
+        f"np.ma.array((A.data > 0) * {MAX_8_BIT_INTENSITY}, mask=False)"
+    )
 
     intensity_co = source_asset_co.copy(
         deep=True,
         update={
             "calc": None,
             "band_count": 1,
-            "data_type": DataType.uint8,
+            "data_type": DataType.uint16,
```
mbertrand, line 305: Not familiar with this dataset yet, curious why the data type needs to change.

app/tasks/raster_tile_set_assets/utils.py
```diff
@@ -150,3 +151,60 @@ async def create_gdaldem_job(
...
+
+async def create_resample_job(
+    dataset: str,
+    version: str,
+    co: PixETLCreationOptions,
+    zoom_level: int,
+    job_name: str,
+    callback: Callback,
+    parents: Optional[List[Job]] = None,
+):
```
mbertrand, line 164: Add a brief explanatory docstring?

batch/python/resample.py
```diff
@@ -0,0 +1,539 @@
...
+            raise Exception("Error creating VRT!")
+
+    return vrt_path
+
+
+def get_s3_client(aws_region=AWS_REGION, endpoint_url=AWS_ENDPOINT_URL):
+    return boto3.client("s3", region_name=aws_region, endpoint_url=endpoint_url)
+
+
+def get_s3_path_parts(s3url) -> Tuple[str, str]:
```
mbertrand, line 136:
```python
def get_s3_path_parts(s3url: str) -> Tuple[str, str]:
```
Also, add a docstring after (and for some of the other functions below without them)?

dmannarino:
> One general question: why did you add this in this repo instead of just allowing an option to directly call GDAL CLI in PixETL? I feel like now we've duplicated responsibilities here, and will have to maintain a whole separate way of processing tiles with GDAL.
> 
> How complicated would be to just add an option maybe somewhere around [here](https://github.com/wri/gfw_pixetl/blob/e50ec4ccc0c71fe4d58d616c4aa916b598b6973a/gfw_pixetl/tiles/raster_src_tile.py#L302) in PixETL to split between reading the tile in-memory or just calling GDAL subprocess? Or would that require like a huge refactor?

They're good points, it's far from an ideal situation.
I did spend a while trying to think through what would be involved in refactoring pixetl to use the command line utils (and took another look after you linked to the code), but it's tricky partly because of how general it (pixetl) is. I'm able to get away with the approach here because I'm only replacing one very specific use case (resampling tiles with no calc function being applied).
I'm also hoping that longer-term we can improve the performance and bug situation with pixetl and eliminate this script (which I admit is a bandaid). However I think that refactoring pixetl (without making a total hack of it) would take longer than we have right now.

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -69,49 +69,29 @@ def generate_8_bit_integrated_calc_string() -> str:
...
+    # day = "(A >= 20000) * (A.data % 10000)"
+    # confidence = "(A.data // 30000)"  # 0 for low confidence, 1 for high
+    #
+    # red = f"({day} / 255)"
+    # green = f"({day} % 255)"
+    # blue = f"(A.data >= 20000) * (({confidence} + 1) * 100 + C.data)"
+    #
+    # alpha = f"(B >= 4) * (B.data & 255)"
+    #
+    # return f"np.ma.array([{red}, {green}, {blue}, {alpha}], mask=False)"
```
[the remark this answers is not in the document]
dmannarino: Yep!

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: Eh, good question. To be honest this is a change left over from when I was attempting to diagnose a problem. I knew that in some cases mixing data types in a VRT is problematic, though in this case it turns out not to matter (there's no problem promoting the uint8 to uint16 when read through the VRT with the other uint16 band). But uint8 takes up less space, so I'll change it back. Thanks for asking!

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -154,6 +154,7 @@ async def raster_tile_cache_asset(
             source_reprojection_parent_jobs,
             max_zoom_resampling=PIXETL_DEFAULT_RESAMPLING,
             max_zoom_calc=max_zoom_calc,
+            use_resampler=max_zoom_calc is None,
```
mbertrand, line 157: Based on slack chat this morning, would it be better to do this instead?
```python
use_resampler=calc is None,
```

or 
```python
use_resampler=(calc is None and max_zoom_calc is None),
```

batch/python/resample.py
```diff
@@ -160,15 +161,92 @@ def download_tiles(args: Tuple[str, str]) -> str:
...
+    else:
+        bucket, source_key = get_s3_path_parts(source_tile_uri)
+        s3_client = get_s3_client()
+        print(f"Downloading {source_tile_uri} to {local_src_file_path}")
+        s3_client.download_file(bucket, source_key, local_src_file_path)
 
     return local_src_file_path
 
 
+def exists_in_s3(target_bucket, target_key):
```
solomon-negusse, line 235: should this go [here](https://github.com/wri/gfw-data-api/blob/master/app/utils/aws.py)?

dmannarino: As things are we don't copy all the data API code into the Batch containers, just the stuff under /batch
But I agree it is general enough that it should be factored-out. Will do so.

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -51,67 +51,44 @@ class SymbologyInfo(NamedTuple):
...
-        "((C.filled(0) >> 1) > 0) * "
-        "(((C.filled(0) & 1) == 0) * 2 + (C.filled(0) & 1))"
+    # The front end is also expecting the confidences of all the alerts in the
+    # original alert systems, packed into a uint16 as the Alpha channel.
+    # Fake the combined confidences because the front end doesn't need to know
+    # which alert systems are what confidence. It REALLY wants to know if the
+    # combined  alert is low, high, or highest confidence. In the future
+    # suggest a new way for the front end to obtain this info that's more
+    # elegant (such as by more efficiently packing into Blue channel)
+    alpha = (
```
jterry64, line 87: Just to understand, you're basically just putting the combined confidence in the alpha  band, and where you'd normally have the bits for the other confidences, it's just empty?

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -51,67 +51,44 @@ class SymbologyInfo(NamedTuple):
...
     red = f"({day} / 255)"
     green = f"({day} % 255)"
-    blue = f"(({confidence} + 1) * 100 + B)"
+    blue = f"(A.data >= 20000) * (({confidence} + 1) * 100 + B.data)"
 
     return f"np.ma.array([{red}, {green}, {blue}])"
 
 
-def generate_8_bit_integrated_calc_string() -> str:
+def integrated_alert_merge_calc() -> str:
```
jterry64, line 66: This methods seems like mostly a copy of the normal date conf encoding function. Would it be better to break it up like this:

- Method to generate red/blue/green channels
- Method to generate alpha channel
- Method that calls RGB method, and also calls if alpha method if necessary, and combined them all into masked array

app/tasks/raster_tile_cache_assets/utils.py
```diff
@@ -80,6 +90,7 @@ async def create_wm_tile_set_job(
     creation_options: RasterTileSetSourceCreationOptions,
     job_name: str,
     parents: Optional[List[Job]] = None,
+    use_resampler: bool = False,
```
jterry64, line 93: Do you need to pass this param through all these functions, or can you just determine whether to resample from the creation options here?

solomon-negusse (approved):

batch/python/resample.py
```diff
@@ -252,44 +303,20 @@ def compress_raster(source_path, target_path):
     if translate_process.returncode < 0:
         raise SubprocessKilledError
     elif translate_process.returncode > 0:
-        print(translate_process.stderr)
+        logger.error(translate_process.stderr)
```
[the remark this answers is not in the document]
solomon-negusse: Thanks for shedding light on how the multiprocess logging works!

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: Yes! With one other detail that isn't clear because I made a Math error: For the highest confidence case it ALSO indicates an alert in one of the other bits (I arbitrarily chose RADD for a single alert and RADD + GLADS2 for multiple). The frontend just checks "Is there an alert? In more than one system? If so: Highest. If not: Confidence is the one indicated in that one system."
I will push a better explanation and fix the Math error.

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: Good point, deduplication done!

app/tasks/raster_tile_cache_assets/utils.py
dmannarino: Mmm, good question. Because we lose a little context it would slightly widen the range of circumstances in which the resampling script would run if for example we just checked "if creation_options.calc is None" in this function. I have to think carefully about what things might get included that it can't handle, or that at least haven't been tested. I keep changing my mind, but would like to err on the side of caution and save this for a future clean-up + testing. I'll add a TODO to that effect.

PR #288, opened 2022-03-30
Hotfix: Add 10km grid; Swap Red and Blue channels to match last year's TCL; allow specifying timeout for RTC jobs

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

- There is no grid suitable for 10km resolution datasets
- Red and Blue channels for year-intensity symbology seem to have been swapped since last year's TCL release
- One cannot specify a timeout for raster tile caches

Issue Number: N/A

## What is the new behavior?

- There is a new 90/1008 grid suitable for 10km resolution datasets
- Red and Blue channels for year-intensity symbology have been made consistent with last year's TCL release
- One CAN specify a timeout for raster tile caches

- [ ] Yes
- [x] No

### DIFF
--- a/app/tasks/raster_tile_set_assets/utils.py
+++ b/app/tasks/raster_tile_set_assets/utils.py
@@ -141,6 +141,10 @@ async def create_gdaldem_job(
         target_prefix,
     ]
 
+    kwargs = dict()
+    if co.timeout_sec is not None:
+        kwargs["attempt_duration_seconds"] = co.timeout_sec
+
     return GDALDEMJob(
         dataset=dataset,
         job_name=job_name,
@@ -148,6 +152,7 @@ async def create_gdaldem_job(
         environment=JOB_ENV,
         callback=callback,
         parents=[parent.job_name for parent in parents] if parents else None,
+        **kwargs,
     )
 
 
@@ -209,6 +214,10 @@ async def create_resample_job(
         target_prefix,
     ]
 
+    kwargs = dict()
+    if co.timeout_sec is not None:
+        kwargs["attempt_duration_seconds"] = co.timeout_sec
+
     return PixETLJob(
         dataset=dataset,
         job_name=job_name,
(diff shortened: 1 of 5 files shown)


jterry64 (approved):

app/tasks/raster_tile_set_assets/utils.py
```diff
@@ -141,13 +141,18 @@ async def create_gdaldem_job(
         target_prefix,
     ]
 
+    kwargs = dict()
```
jterry64, line 144: minor, but is there a way to not use kwargs for this? I think we overuse submitting params as kwargs sometimes, and it gets confusing what's being used where. I think we could just set a default attempt_duration_seconds if not specified (or leave it None)

jterry64 (approved):

app/tasks/raster_tile_set_assets/utils.py
dmannarino: Hmm. There IS a default attempt_duration_seconds, but it's specified in the model. The problem with passing in a None is that it then won't take the default value (as specified in the Job model). It will wind-up as REALLY None, when we want it to be the default if unspecified:
In [9]: class Blah(BaseModel):
   ...:     foo: Optional[int] = 5
In [10]: some_blah = Blah(foo=None)
In [11]: type(some_blah.foo)
Out[11]: NoneType
There IS a default

tanderegg: One option would be to move all the params to a dict called job_params (or whatever), and then just do `return GDALDEMJob((**job_params)`, which might be a bit more clear than mixing **kwargs with named arguements?  That way all parameters are defined together, some of them conditional, before being passed all together to the constructor.

dmannarino: Mmm, that's an idea. How do you feel about that Justin?

PR #307, opened 2022-06-09
GTC 1862 spike pagination

gtempus:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): This is a spike to learn how to use the [fastapi-pagination](https://github.com/uriyyo/fastapi-pagination) package

No pagination on the `datasets` endpoint
Issue Number: N/A

## What is the new behavior?

Pagination is now happening on the test endpoint: `/datasets?page=1&size=10` 
-
-
-

- [ ] Yes
- [x] No

## Other information

This PR is not intended for merging. It's just an experiment to learn about fastapi-paginate.
However, I still wanted to avoid breaking the existing functionality. Therefore, I did add an extra route and small function to the `crud/datasets` module.

The response model is not sufficient for production. It was just a way to get something coming back from the server.

Manual Testing:
- Original behavior endpoint: `/datasets`
  - The result is unchanged. No new pagination fields are included.
- To enable pagination: `/datasets?page=1&size=1`
  - `page` needs to be greater than `0`
  - `size` can be between `1` and `49`, inclusive*
  - Now there are `links` and `meta` properties included in the response.
*I have some silly logic right now that decides to paginate whether `size < 50`

Initial thoughts:
- SQLAlchemy - The install of fastapi-pagination did cause the update of several packages. I had to introduce tighter constraints on `sqlalchemy` and pin it to `1.3.x` because it seemed that `1.4.x` was breaking the API.
- We'll need a way to get to the SQLAlchemy Query object before causing gino to execute the query. It's not a dealbreaker but some refactoring of the crud modules will need to happen for the proposed endpoints.
- Some consideration will need to go into extending the Page and/or DatasetsResponse types to bring them into alignment with Resource Watch. Extension/Customization steps are outlined in the [fastapi-pagination repo](https://github.com/uriyyo/fastapi-pagination)


jterry64:
@gtempus I don't foresee any breaking changes with the package updates, as long as tests are passing.

This is cool! Do you know why the response has the __value__ and __profile__ keys for each object, and is that something we can customize? Seems like it might it be like Python hidden attribute names based on the underscores?

And yeah, if we're going to use the page number/size method, we should probably align with RW. 

I haven't implemented pagination on anything in awhile, but based on the amount of customization we might need to do for this library, would it be worth it vs just implementing pagination ourselves? I.e., does this library provide good value?

Those are just my initial thoughts.s

app/routes/datasets/datasets.py
```diff
@@ -16,9 +17,17 @@
...
+        query = datasets.get_datasets_query()
+        data = await paginate(query, params)
+        adapted = [dataset.__values__ for dataset in data.items]
+        return DatasetsResponse(data=adapted,
+                                    links={'self': "", 'first': "", 'last': "", 'prev': "", 'next': ""},
+                                    meta={'size': data.size, 'total-pages': "", 'total-items': data.total},
+                                    status='success')
 
     data = await datasets.get_datasets()
+    return DatasetsResponse(data=data, status='success')
```
gtempus, line 32: This actually turned out to be the hardest part for introducing a breaking changes _without_ actually breaking things. The main idea was to have the same endpoint _do the right thing_ if pagination params were given but otherwise remain unchanged to preserve legacy behavior.

With `pydantic` I couldn't just create subclasses of `DatasetsResponse` and continue to return the generalized one. When I did actually tried to do that, `pydantic` would throw a ValidationError and tell me that I was either including extra fields or it would add the pagination properties when they should've been hidden.

app/routes/datasets/datasets.py
```diff
@@ -16,9 +17,17 @@
...
-async def get_datasets() -> DatasetsResponse:
+async def get_datasets(params: Params = Depends()) -> DatasetsResponse:
     """Get list of all datasets."""
+    if params.size < 50:
+        query = datasets.get_datasets_query()
+        data = await paginate(query, params)
+        adapted = [dataset.__values__ for dataset in data.items]
+        return DatasetsResponse(data=adapted,
+                                    links={'self': "", 'first': "", 'last': "", 'prev': "", 'next': ""},
+                                    meta={'size': data.size, 'total-pages': "", 'total-items': data.total},
```
gtempus, line 28: Pydantic types for these would be created in the actual implementation.

app/models/pydantic/responses.py
```diff
@@ -1,8 +1,75 @@
...
+            return dict_orig(*args, **kwargs, exclude_unset=exclude_unset)
+        model.dict = dict_excludes_unset
+
+        # Done
+        return model
+
+    return decorator
+
+
+@partial('links', 'meta')
```
gtempus, line 72: I did some research to see how other folks were solving similar issues of how to get `pydantic` to completely remove various properties. My implementation is a working refinement of the [thoughts in this pydantic thread](https://github.com/samuelcolvin/pydantic/issues/1223). 

It's a decorator that lets me define properties that should not exists (in any way) if they are not explicitly passed in during initialization. That way, there are no new properties revealed when legacy list endpoints are called (so we don't inadvertently break existing clients). 

It's not perfect. Because of the `exclude_unset` directive, we would need to pass the `success` property explicitly now from the route code. I don't think that's a deal breaker.

gtempus (commented):
After walking though the code with @dmannarino we made these decisions:
### Response Types
- Use `Union` to explicitly show that one of two types of responses could happen (paginated or _not_ paginated for legacy support)
- Make `pydantic` types for the `links` and `meta` properties within the paginated response.
### Pagination Implementation
- Use [fastapi-pagination](https://github.com/uriyyo/fastapi-pagination)
- Design our usage around it so that we can easily replace it, if necessary.
- Spend some effort separating query building from execution so that pagination can be injected cleanly into the existing crud modules
### Documentation
- Ensure that any pagination information is clear in the API docs

PR #309, opened 2022-06-15
GTC-1877: Check all geostores; add AMD instance types; update deps

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Clients have to know where a given geostore is located (gfw or rw) to use it
- We return a 500 instead of a 404 if a geostore is not found in the RW API
- We don't make use of AMD instances

Issue Number: GTC-1877

## What is the new behavior?

- Clients DO NOT have to know where a given geostore is located (gfw or rw) to use it (all origins are checked)
- We return a 404 if a geostore is not found in either locations (gfw or rw)
- We DO make use of AMD instances

- [x] Yes
- [x] No

## Other information

I had to update dependencies (and therefore a number of tests) because I added the respx library to mock HTTPX calls
I also gratuitously updated the versions of libraries used by the pre-commit hook


app/utils/geostore.py
```diff
@@ -51,3 +73,54 @@ async def get_geostore(geostore_id: UUID, geostore_origin: str) -> GeostoreCommo
...
+    geostore_constructor = {
+        GeostoreOrigin.gfw: _get_gfw_geostore,
+        GeostoreOrigin.rw: rw_api.get_geostore,
+    }
+
+    geo_func = geostore_constructor.pop(geostore_origin)
+
+    # If we get a geostore from any origin return it,
+    # if we get all 404s return a 404,
+    # if we get a mixture of errors return a 500 with explanation
```
gtempus, line 92: Nice! Does this need to make it into the rdocs, @dmannarino?

app/utils/geostore.py
```diff
@@ -51,3 +73,54 @@ async def get_geostore(geostore_id: UUID, geostore_origin: str) -> GeostoreCommo
...
+            exceptions.append(e)
+        except Exception as e:
+            logger.exception(e)
+            exceptions.append(e)
+
+    non_404s = [
+        exception
+        for exception in exceptions
+        if not isinstance(exception, RecordNotFoundError)
+    ]
```
gtempus, line 117: Ah, nice filter.

app/utils/geostore.py
```diff
@@ -51,3 +73,54 @@ async def get_geostore(geostore_id: UUID, geostore_origin: str) -> GeostoreCommo
...
+
+    if non_404s:
+        msg = (
+            f"One or more errors were encountered looking for geostore "
+            f"{geostore_id}. Please email info@wri.org for help."
+        )
```
gtempus, line 123: I like this. So, who monitors `info@wri.org`?

gtempus (changes requested):
This was easy to read through, @dmannarino! I just left a couple questions.

app/utils/geostore.py
dmannarino: A good question. I am no longer sure this is the best address to use. data@wri.org would probably be better, I think that's monitored by Logan. But I'll ask him. I thought there was one monitored by Maanas, who is perhaps the best. For the moment I'll set it to mine, as it's the only gentlemanly thing to do in doubt.

app/utils/geostore.py
dmannarino: Mmm, good point. Will add it to the route documentation.

app/utils/geostore.py
gtempus: Ah, okay. Yeah, we should definitely find the _official_ email for this.

terraform/api_gateway.tf
```diff
@@ -46,7 +46,7 @@ module "query_get" {
   authorizer_id = aws_api_gateway_authorizer.api_key.id
   api_resource = module.query_resource.aws_api_gateway_resource
 
-  require_api_key = true
+  require_api_key = false
```
solomon-negusse, line 49: Will need to disable api keys in the backend by commenting out [this](https://github.com/wri/gfw-data-api/blob/2eb041b99ed3af0ac9ad7c954de50561c3650391/app/routes/datasets/queries.py#L124) and the remaining endpoints in the module as well.

app/settings/globals.py
```diff
@@ -179,3 +179,7 @@
 GOOGLE_APPLICATION_CREDENTIALS = config(
     "GOOGLE_APPLICATION_CREDENTIALS", cast=str, default="/root/.gcs/private_key.json"
 )
+
+FEATURE_CHECK_ALL_GEOSTORES = config(
```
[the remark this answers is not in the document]
jterry64: If this is the right implementation, should the feature flags be in a dictionary of all feature flags instead of each getting their own global var?

app/utils/geostore.py
```diff
@@ -34,7 +40,23 @@ async def _get_gfw_geostore(geostore_id: UUID) -> GeostoreCommon:
     return geostore_common
 
 
-async def get_geostore(geostore_id: UUID, geostore_origin: str) -> GeostoreCommon:
+def check_all_geostores():
```
jterry64, line 43: Similar to my comment above, do we want to be making a separate function like this for every feature flag, or just one function where we pass the feature flag name?

app/utils/geostore.py
```diff
@@ -34,7 +40,23 @@ async def _get_gfw_geostore(geostore_id: UUID) -> GeostoreCommon:
...
+async def get_geostore(
+    geostore_id: UUID, geostore_origin: GeostoreOrigin
+) -> GeostoreCommon:
+    if check_all_geostores():
+        return await get_geostore_from_any_origin(geostore_id, geostore_origin)
+    else:
+        return await get_geostore_legacy(geostore_id, geostore_origin)
+
+
+async def get_geostore_legacy(
```
jterry64, line 56: Instead of having a function named legacy and function that gets from any origin, could we just have one function that accepts a list of origins and searches for them in that order? For the legacy function, just only send one origin.

tests/routes/datasets/test_versions.py
```diff
@@ -80,12 +81,12 @@ async def test_versions(async_client):
     response = await async_client.patch(
         f"/dataset/{dataset}/{version}", json={"is_latest": True}
     )
-    print(response.json())
+    # print(response.json())
```
tanderegg, line 84: Here too, might as well just remove `print`'s completely.

app/settings/globals.py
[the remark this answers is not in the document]
gtempus: Good point, @tanderegg. We should also have a JIRA card that reminds us to remove the temporary toggle. That'll keep our feature list from growing, too.

tests_v2/unit/app/routes/datasets/test_queries.py
```diff
@@ -36,9 +36,10 @@ async def test_query_dataset_with_api_key(
         f"/dataset/{dataset_name}/{version_name}/query",
         params=params,
         headers=headers,
+        follow_redirects=True,
     )
 
-    print(response.json())
+    # print(response.json())
```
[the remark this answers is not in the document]
gtempus: Good catch, @tanderegg 👍

gtempus:
I love all the discussion in these reviews! ❤️

app/settings/globals.py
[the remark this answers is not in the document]
gtempus: That's sounds great, @jterry64!

PR #313, opened 2022-06-23
Metadata models

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Entity (dataset|version|asset) metadata are stored as JSONB in entity tables so fields aren't validated
- Version and asset metadata include parent metadata
- Metadata includes some fields that have don't get used/populated or can be constructed from other fields

Issue Number: [GTC-1347](https://gfw.atlassian.net/browse/GTC-1347)

## What is the new behavior?

- Move metadata to separate tables to improve data validation, performance, flexibility (e.g., can easily add metadata versioning in future)
- Remove metadata inheritance and reduce redundancy
  - assets metadata don't include dataset/version metadata (gives better alignment with STAC as well)
  - reduce overlapping fields between dataset and version metadata
  - remove unused fields (looking at production data)
- Added new fields (informed by STAC)

- [x] Yes
- [ ] No

Platforms are still using the legacy metadata so won't be impacted by this but will to migrate from legacy to this metadata.


app/models/pydantic/assets.py
```diff
@@ -19,7 +19,7 @@ class Asset(BaseRecord):
     status: AssetStatus = AssetStatus.pending
     is_managed: bool
     is_downloadable: bool
-    metadata: AssetMetadata
+    metadata: Optional[Union[AssetMetadataOut, BaseModel]]
```
dmannarino, line 22: I forget, why do you need to allow Basemodel? I think I saw why elsewhere in the PR but I can't find it ATM.

app/models/orm/asset_metadata.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    pixel_meaning = db.Column(db.String, primary_key=True)
+    description = db.Column(db.String)
+    alias = db.Column(db.String)
+    data_type = db.Column(db.String)
+    unit = db.Column(db.String)
+    compression = db.Column(db.String)
+    no_data_value = db.Column(db.String)
+    statistics = db.Column(db.JSONB)
+    values_table = db.Column(db.ARRAY(db.JSONB))
```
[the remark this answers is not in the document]
solomon-negusse: For datasets with discrete values such as land cover/land use or deforestation alert confidence, this maps the pixel values to what they mean.

app/models/pydantic/assets.py
solomon-negusse: This is to allow returning empty metadata responses since we allow creating assets without associated metadata.

app/models/pydantic/metadata.py
```diff
@@ -1,176 +1,186 @@
...
 
-class RasterTableRow(StrictBaseModel):
-    value: int
-    meaning: Any
+    class Config:
+        schema_extra = {
+            "examples": [
+                {
+                    "content_date_range": {
+                        "start_date": "2000-01-01",  # TODO fix date
```
dmannarino, line 127: Is this TODO still something you want to fix? (not sure what is wrong with it)

dmannarino:
Last question: In the cases where you inherit from BaseModel, could you inherit from StrictBaseModel instead? We've been doing that in order to generate an error when someone tries to manipulate an object in a way that breaks the schema.

solomon-negusse:
> Last question: In the cases where you inherit from BaseModel, could you inherit from StrictBaseModel instead? We've been doing that in order to generate an error when someone tries to manipulate an object in a way that breaks the schema.

Good catch @dmannarino! I have updated `models/pydantic/metadata` reverting to `StrictBaseModel`. The remaining instances of `BaseModel` use have to do with CRUD on entities where we allow empty metadata for cases when users may not have that ready when registering a resource (e.g. dataset[https://github.com/wri/gfw-data-api/blob/metadata-models/app/models/pydantic/datasets.py#L13](url)) but the `Union` type in this case is smart enough to disallow non-empty metadata objects with invalid properties.

dmannarino:
Is there a significance to the fact that the migration appears under "draft_versions"? Is that a problem?

dmannarino (approved):

solomon-negusse:
> Is there a significance to the fact that the migration appears under "draft_versions"? Is that a problem?

@dmannarino, I wanted the migration script to be backed up on GH and have consistent commit history as rest of code so I put in a folder where alembic isn't going to run it.

PR #321, opened 2022-08-10
Base symbology assets on WM asset, not original EPSG:4326 asset

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- Symbology assets are made based on the original asset, rather than on the WM one, taking extra time (and introducing possibility for discrepancy between pixetl and resample script)

Issue Number: GTC-1924

## What is the new behavior?

- Symbology-created assets are generated based on the WM asset rather than the original source asset

- [ ] Yes
- [x] No


app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -65,7 +65,7 @@ async def raster_tile_cache_asset(
                 AssetType.raster_tile_set,
                 source_asset.creation_options,
             )
-        ).replace("/geotiff", "/gdal-geotiff")
```
gtempus, line 68: This `replace` looks like it was tacked on, originally. Now that it's being removed, what assets _could_ be affected?

app/tasks/raster_tile_cache_assets/symbology.py
```diff
@@ -548,7 +548,7 @@ async def _create_intensity_asset(
     )
 
     if zoom_level == max_zoom:
-        parent_jobs: Optional[List[Job]] = None
+        parent_jobs: List[Job] = [jobs_dict[zoom_level]["source_reprojection_job"]]
```
gtempus, line 551: I haven't spent much time looking at `tasks`, yet. Is this how we create processing pipelines? Does this schedule work to be done or describe a dependency for the child jobs? 🤷

gtempus (commented):
I just have some questions since this is a new part of the repo for me, @dmannarino. Thanks!

app/tasks/raster_tile_cache_assets/symbology.py
dmannarino: It describes a dependency for the child jobs.
We probably could handle the dependencies a little better than a dictionary. For the moment, look at https://github.com/wri/gfw-data-api/blob/upgrade_packages_tests/app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py#L120-L180 to better understand what's happening. 
To see where the actual submission to AWS Batch happens, look at [gfw-data-api/app/tasks/batch.py](https://github.com/wri/gfw-data-api/blob/upgrade_packages_tests/app/tasks/batch.py)

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Yeah, I had originally slipped it in as an aspirational optimization (i.e. I have no idea if it actually made a difference). I am trying to revert it now because I feel self-conscious about tacking it on, and I haven't entirely thought through whether or not it will always work (I think it does by luck). The change should not affect any existing assets, though, only new ones which are created.

app/models/orm/migrations/env.py
```diff
@@ -98,7 +98,7 @@ def run_migrations_online():
         with context.begin_transaction() as transaction:
             context.run_migrations()
             if "dry-run" in context.get_x_argument():
-                print("Dry-run succeeded; now rolling back transaction")
```
jterry64, line 101: Is this meant to be here?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
jterry64: Just so I understand, when creating a tile cache, it tries to use the gdal-geotiff assets, but you're changing it to just use the normal geotiff assets because the WM assets don't have gdal-geotiff assets?

app/models/orm/migrations/migrate_asset_tables.py
```diff
@@ -0,0 +1,91 @@
+"""Add version column to all data tables."""
```
jterry64, line 1: All this migration stuff is just for revisions?

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -150,7 +150,9 @@ async def raster_tile_cache_asset(
         symbology_jobs: List[Job]
         symbology_uri: str
 
-        symbology_co = source_asset_co.copy(deep=True)
+        symbology_co = source_asset_co.copy(
```
jterry64, line 153: Could you add a comment or move this out into a function with a clearer name of what's happening? Had to do dig through the  code to understand why this made sense.

jterry64 (approved):
Just minor comments, also I think we need to find a better way to deal with migrations in dev... I spent awhile trying to figure out why you're doing all this migration work only to realize it's just there to make run in dev.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: > Just so I understand, when creating a tile cache, it tries to use the gdal-geotiff assets, but you're changing it to just use the normal geotiff assets because the WM assets don't have gdal-geotiff assets?
Yeah, we don't create the gdal-geotiff assets with the resample script, and if there's any benefit to doing so it's minimal. And this only changed to using gdal-geotiffs in one isolated instance. I think better to be consistent unless we adopt using gdal-geotiffs comprehensively.

app/models/orm/migrations/env.py
dmannarino: As I think you subsequently realized, this will disappear when I revert the dev-appeasing migration stuff.

app/models/orm/migrations/migrate_asset_tables.py
dmannarino: Yup.

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
dmannarino: Will do.

app/tasks/raster_tile_cache_assets/symbology.py
gtempus: Thanks for the links, @dmannarino!

PR #354, opened 2022-11-15
Gtc-2131: Implement vector appends

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- Appends to vector assets are not supported
- Vector source creation is not very retryable

Issue Number: GTC-2131

## What is the new behavior?

- Appends to vector assets are supported
- Vector source creation is significantly more retryable
- Lots of new tests
- Some reorganization of existing tests, code
- SQL snippets are factored-out to reduce error-prone duplication
- moto dependency now finally back to following upstream

- [ ] Yes
- [x] No

## Other information

Dev testing notes:
Performed a PUT to `http://gfw-data-api-elb-gtc-2131-vector-2029840344.us-east-1.elb.amazonaws.com/dataset/wdpa_protected_areas/v202208.1` with the following JSON payload:
{
    "creation_options": {
        "source_type": "vector",
        "source_driver": "FileGDB",
        "source_uri": [
            "s3://gfw-data-lake-dev/wdpa_protected_areas/v202208/raw/WDPA_WDOECM_Sep2022_Public_all.gdb.zip"
        ],
        "layers": [
            "WDPA_WDOECM_poly_Sep2022_all"
        ],
        "create_dynamic_vector_tile_cache": true,
        "add_to_geostore": true,
    }
}

More test details to come, currently blocked on migration issues in dev.

[GTC-2131]: https://gfw.atlassian.net/browse/GTC-2131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/pydantic/creation_options.py
```diff
@@ -191,18 +191,50 @@ class VectorSourceCreationOptions(StrictBaseModel):
...
+            assert len(v) >= 1, "CSV sources require at least one input file"
+        else:
+            assert (
+                len(v) == 1
+            ), "Non-CSV vector sources require one and only one input file"
+        return v
+
+
+class VectorSourceAppendOptions(StrictBaseModel):
+    source_type: VectorSourceType = Field(..., description="Source type of input file.")
```
jterry64, line 217: In the tabular append, I think we just accept source URIs and force it to use the same input options (i.e. be in the same format) as the data on initial creation. 

Does it make sense to allow potentially different formats go into the same table? I guess as long as the schema is the same it should work.

batch/scripts/create_vector_schema.sh
```diff
@@ -8,11 +8,20 @@ set -e
...
+# optional arguments
+# -g | --geometry_name (get_arguments.sh specifies default)
+# -i | --fid_name (get_arguments.sh specifies default)
+
 ME=$(basename "$0")
 . get_arguments.sh "$@"
 
+set -u
+
+# FIXME: Downloading the whole file for this step might be unnecessary
```
jterry64, line 23: I think you would need to unless it's CSV? You would need to the whole shapefile or geodatabase downloaded to access its schema.

batch/scripts/load_vector_csv_data.sh
```diff
@@ -6,22 +6,51 @@ set -e
...
-# -f | --local_file
-# -X | --zipped
+
+# optional arguments
+# -g | --geometry_name (get_arguments.sh specifies default)
 
 ME=$(basename "$0")
 . get_arguments.sh "$@"
 
+set -u
```
jterry64, line 16: So presumably this will happen now whether it's table creation or an append? Have you tested this on a very large table to see if it runs into memory issues? (IIRC the temp tables are just stored in-memory) Since we break up the load data calls into smaller jobs, I think this shouldn't be an issue?

app/models/pydantic/creation_options.py
jterry64: And on the same note, should we actually be enforcing somewhere that the schema is the same as the original table, and return a 4XX error if not? Presumably it'll just fail later down the pipeline and you'd get a vague message in the task log. 

I doubt we check that now for tabular appends, and it'd be tricky since we only have a batch script for checking the schema (and may need to download the whole file for certain formats). So maybe this is more of a nice-to-have for later.

batch/scripts/_add_gfw_fields_sql.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
jterry64, line 1: I really like that you separated these out into subscripts, but I wonder if there's an established pattern or naming convention for doing that with bash scripts? The "_" in front of the script feels more like a python thing, although works

jterry64: A quick googling makes me think we have a pretty uncommon use case of using bash scripts for an actual processing pipeline instead of one-off system admin scripts... so maybe just do as you like!

jterry64 (changes requested):
Mostly questions more than change requests, looks pretty good!

solomon-negusse (commented):
Looks good to me.  The choice of fixed `temp_table` made me wonder though if there's a potential for conflict if there are concurrent jobs running. Just [read](https://www.postgresql.org/docs/current/sql-createtable.html) that postgres puts temporary tables in a special schema and not clear to me if they are all placed in a single schema.

dmannarino:
> Looks good to me. The choice of fixed `temp_table` made me wonder though if there's a potential for conflict if there are concurrent jobs running. Just [read](https://www.postgresql.org/docs/current/sql-createtable.html) that postgres puts temporary tables in a special schema and not clear to me if they are all placed in a single schema.

The temporary table is supposedly inaccessible from outside that same transaction, so I'm PRETTY sure that concurrent jobs can run without interfering... On the other hand it should be harmless and easy enough to make the table name unique. So I'll do that.

batch/scripts/load_vector_csv_data.sh
dmannarino: As mentioned in Slack, this approach works with WDPA once upgrading the dev writer to the same instance class as staging/production.

app/models/pydantic/creation_options.py
dmannarino: > In the tabular append, I think we just accept source URIs and force it to use the same input options (i.e. be in the same format) as the data on initial creation.
> 
> Does it make sense to allow potentially different formats go into the same table? I guess as long as the schema is the same it should work.

Personally I'd like to allow things unless there's a specific reason not to. On the "for" side, I can imagine a situation in which updates start to be delivered in a different format than the original data at some point. But mostly I just dislike the idea of disallowing it without a strong reason to.

dmannarino: > And on the same note, should we actually be enforcing somewhere that the schema is the same as the original table, and return a 4XX error if not? Presumably it'll just fail later down the pipeline and you'd get a vague message in the task log.
> 
> I doubt we check that now for tabular appends, and it'd be tricky since we only have a batch script for checking the schema (and may need to download the whole file for certain formats). So maybe this is more of a nice-to-have for later.

Mmm, good point. Since it's possible to specify the schema, we could require that if they specify in the original they need to specify in the append, and that they should match.
But you're right, it sounds like checking the actual schema of the data would at least sometimes (often?) require downloading the entire data file, which at least sometimes (often?) would cause a timeout.
Let me check, though. Maybe ogrinfo has some format-specific magic that can determine the schema of these files remotely without downloading the entire files. I'll investigate.

dmannarino: On the other hand, requiring them to re-specify the schema when we don't really need/use it seems like it's adding an annoying hurdle. What do you think, @jterry64 ?

dmannarino: So, plot twist: The way I wrote it didn't work. The append endpoint expects a JSON payload of (only)
{
  "source_uri": [some_uri(s)]
}
In order to support specifying a schema and such I would have to either make a breaking change to the API to or a special case for vector appends, which would be nasty. I don't like it, but I've made the vector append case consistent with the table append case. It's something I'd like to change when we increment API versions, though.

batch/scripts/create_vector_schema.sh
dmannarino: Good news! ogrinfo seems to be able to get the schema from some remote file types! I think this change could be back on the table (for the future).

app/tasks/utils.py
```diff
@@ -1,13 +1,25 @@
...
-        self._lists = [list() for i in range(0, size)]
-        self._idx = -1
+    """A data structure that consists of a number of lists attached to a
+    circular buffer.
+
+    One may iterate over it and append items to the element in hand in
+    order to evenly distribute whatever it is one has amongst the
+    different lists. Later, one can call the all() method to get all the
+    lists to do with as one pleases.
+    """
```
gtempus, line 17: Thanks for the docstring! 🙇

app/utils/path.py
```diff
@@ -21,18 +21,16 @@ def infer_srid_from_grid(grid: str) -> str:
...
-    """Get basename of source file.
-
-    If Zipfile, add VSIZIP prefix for GDAL
-    """
-    bucket, key = split_s3_path(s3_uri)
-    client = get_s3_client()
+    """Determine from file name or header if URI points to a Zip file."""
     _, ext = os.path.splitext(s3_uri)
+    if ext.lower() == ".zip":
+        return True
```
gtempus, line 27: I like this early `return`

batch/scripts/_add_gfw_fields_sql.sh
gtempus: Yeah, breaking these up was a really good idea! 🔨

batch/scripts/create_vector_schema.sh
gtempus: Nice! Did you happen to create a Jira card for this `FIXME`? 😉

scripts/test_v2
```diff
@@ -24,6 +24,12 @@ do
       SLOW=--with-slow-tests
       shift # past argument
       ;;
+      --moto-port=*)
+      # prevent port binding issues by explicitly setting the motoserver s3 port
+      # https://developer.apple.com/forums/thread/682332
+      export MOTO_PORT="${key#*=}"
+      shift # past argument=value
+      ;;
```
gtempus, line 32: Thank you for adding this, @dmannarino. I'm not sure why I didn't the first time around.... 🤷

tests/tasks/test_vector_source_assets.py
```diff
@@ -131,75 +132,6 @@ async def test_vector_source_asset(batch_client, async_client):
...
-        )
-        assert response.status_code == 400
-
-        # Downloads
-
-        response = await async_client.get(
-            f"/dataset/{dataset}/{version}/download/csv?sql=select count(*) from mytable;",
-            follow_redirects=True,
-        )
-        assert response.status_code == 200
```
gtempus, line 200: Thank you for splitting these out! 🪓  😅

tests_v2/conftest.py
```diff
@@ -66,8 +67,7 @@ def init_db():
         yield
 
 
-@pytest.fixture()
-@pytest.mark.asyncio
+@pytest_asyncio.fixture
```
gtempus, line 70: 💯

tests_v2/unit/app/tasks/test_vector_source_assets.py
```diff
@@ -0,0 +1,575 @@
...
+
+async def dummy_function():
+    pass
+
+
+async def mock_callback(task_id: UUID, change_log: ChangeLog):
+    return dummy_function
+
+
+class TestVectorSourceAssetsHelpers:
```
gtempus, line 51: It's great to have all these new tests! 🎉

gtempus (approved):
Thanks a ton for walking me through the sizable PR, @dmannarino! I know it much better now! 🙇

batch/scripts/create_vector_schema.sh
dmannarino: I have now! https://gfw.atlassian.net/browse/GTC-2233

app/routes/datasets/versions.py
```diff
@@ -111,6 +111,10 @@ async def add_new_version(
 
     _verify_source_file_access(creation_options["source_uri"])
 
+    # TODO: Do more to verify that any specified options are valid for
+    #  the actual source file. For example, check any specified schema
+    #  with ogrinfo for vector files
```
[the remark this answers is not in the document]
dmannarino: Sounds good, will do! Thanks!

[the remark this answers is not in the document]
dmannarino: Here's this one: https://gfw.atlassian.net/browse/GTC-2234

PR #356, opened 2022-11-30
Upgrade python to 3.10

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): 
Tech debt

We're using Python 3.8

Issue Number: GTC-1994

## What is the new behavior?

-Python 3.10 is used

- [ ] Yes
- [x] No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
-FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8-slim
+FROM tiangolo/uvicorn-gunicorn-fastapi:python3.10-slim
 
 # Optional build argument for different environments
 ARG ENV
 
 RUN apt-get update -y \
-    && apt-get install --no-install-recommends -y gcc libc-dev musl-dev postgresql-client libpq-dev
+    && apt-get install --no-install-recommends -y gcc libc-dev musl-dev postgresql-client libpq-dev make
 
-RUN pip install --upgrade pip && pip install pipenv
+RUN pip install --upgrade pip && pip install pipenv==v2022.11.30
 
 # Install python dependencies
 # Install everything for dev and test otherwise just core dependencies
--- a/Pipfile
+++ b/Pipfile
@@ -55,7 +55,7 @@ typing-extensions = "*"
 uvicorn = {version = "*", extras = ["standard"]}
 
 [requires]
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8-slim
+FROM tiangolo/uvicorn-gunicorn-fastapi:python3.10-slim
 
 # Optional build argument for different environments
 ARG ENV
 
 RUN apt-get update -y \
-    && apt-get install --no-install-recommends -y gcc libc-dev musl-dev postgresql-client libpq-dev
+    && apt-get install --no-install-recommends -y gcc libc-dev musl-dev postgresql-client libpq-dev make
```
gtempus, line 7: Why did we have to add `make`?

Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 
-RUN pip install --upgrade pip && pip install pipenv
+RUN pip install --upgrade pip && pip install pipenv==v2022.11.30
```
gtempus, line 9: Why did we have to pin `pipenv`?

gtempus (changes requested):
Thanks, @dmannarino! I just had a couple of questions about the dependencies.

Dockerfile
dmannarino: Building the image failed due to one of the Python packages failing to compile, complaining that make couldn't be found. Adding this fixed it. It seems that the newer Docker image has different packages included by default (I think it uses a newer version of Debian), which is no big deal.

gtempus: Gotcha. Yeah, I thought we talked about this before with another image. I had deja vu when posting the question. 😆 
Thanks!

Dockerfile
dmannarino: The version of pipenv from the day before had a bug that prevented the lockfile being generated, so when I finally got it to work when the new version came out I decided to pin it. pipenv just doesn't have a great track record WRT bugs, so I decided pinning it is better than having builds start failing at some point in the future.

gtempus (approved):

PR #391, opened 2023-05-17
Feature/viirs metadata

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: [GTC-2390](https://gfw.atlassian.net/browse/GTC-2390)
Data-pump updates don't have a way set content_date_range when appending data

## What is the new behavior?

- Add ability to update metadata when appending

- [ ] Yes
- [x] No

## Other information

[GTC-2390]: https://gfw.atlassian.net/browse/GTC-2390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/crud/assets.py
+++ b/app/crud/assets.py
@@ -11,6 +11,7 @@ from app.crud.metadata import (
     update_asset_metadata,
 )
 from sqlalchemy import func
+from async_lru import alru_cache
 
 from ..errors import RecordAlreadyExistsError, RecordNotFoundError
 from ..models.enum.assets import AssetType
@@ -18,6 +19,7 @@ from ..models.orm.asset_metadata import AssetMetadata as ORMAssetMetadata
 from ..models.orm.assets import Asset as ORMAsset
 from ..models.orm.versions import Version as ORMVersion
 from ..models.pydantic.creation_options import CreationOptions, creation_option_factory
+from ..models.pydantic.asset_metadata import RasterTileSetMetadataOut
 from . import update_data, versions
 
 
@@ -35,7 +37,7 @@ async def get_assets(dataset: str, version: str) -> List[ORMAsset]:
 
     return rows
 
-
+@alru_cache(maxsize=128)
 async def get_raster_tile_sets():
     latest_tile_sets = await (
         ORMAsset.join(ORMVersion)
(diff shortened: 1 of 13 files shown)


jterry64 (approved):
Didn't even think about that, looks good to me! I guess alternatively the data API could just directly PATCH the metadata after appending. The only case I can see where that's better is if the append task fails, the content date range would already be updated, so clients like flagship would think there's new data when there isn't. @solomon-negusse @dmannarino what do you think?

dmannarino (approved):

solomon-negusse:
> Didn't even think about that, looks good to me! I guess alternatively the data API could just directly PATCH the metadata after appending. The only case I can see where that's better is if the append task fails, the content date range would already be updated, so clients like flagship would think there's new data when there isn't. @solomon-negusse @dmannarino what do you think?

I actually started out implementing that but without persisting the metadata payload, don't see a way of passing that to where jobs statuses are [checked](https://github.com/wri/gfw-data-api/blob/master/app/routes/tasks/task.py#L190) and updated. I resorted to this way as the version status will be set to failed if append fails anyway (so client shouldn't rely on the content date range). That will need manual intervention to revert to valid content date range and not too happy about the solution.
Does data-pump check if an append succeeds? In that case, can explore sending the `PATCH` request from data-pump on append success. I'm assuming moving the content end date around would be easier in data-pump than in Data API but need to validate that.

dmannarino:
> Didn't even think about that, looks good to me! I guess alternatively the data API could just directly PATCH the metadata after appending. The only case I can see where that's better is if the append task fails, the content date range would already be updated, so clients like flagship would think there's new data when there isn't. @solomon-negusse @dmannarino what do you think?

Hmm, that's a good point. TBH I favor PATCHing the metadata after the append succeeds, like is done with the metadata and is_latest for the deforestation alerts. But I wouldn't object to this.

solomon-negusse:
> > Didn't even think about that, looks good to me! I guess alternatively the data API could just directly PATCH the metadata after appending. The only case I can see where that's better is if the append task fails, the content date range would already be updated, so clients like flagship would think there's new data when there isn't. @solomon-negusse @dmannarino what do you think?
> 
> I actually started out implementing that but without persisting the metadata payload, don't see a way of passing that to where jobs statuses are [checked](https://github.com/wri/gfw-data-api/blob/master/app/routes/tasks/task.py#L190) and updated. I resorted to this way as the version status will be set to failed if append fails anyway (so client shouldn't rely on the content date range). That will need manual intervention to revert to valid content date range and not too happy about the solution. Does data-pump check if an append succeeds? In that case, can explore sending the `PATCH` request from data-pump on append success. I'm assuming moving the content end date around would be easier in data-pump than in Data API but need to validate that.

Turns out data-pump does check if append succeds, so I'm sending the PATCH request from that side so reverts the relevant changes here.

PR #420, opened 2023-09-05
Fix a tiny typing error

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/app/tasks/dynamic_vector_tile_cache_assets.py
+++ b/app/tasks/dynamic_vector_tile_cache_assets.py
@@ -7,7 +7,7 @@ from ..crud import assets
 from ..models.enum.assets import AssetStatus, is_database_asset
 from ..models.enum.change_log import ChangeLogStatus
 from ..models.enum.creation_options import IndexType
-from ..models.orm.asset_metadata import FieldMetadata
+from ..models.orm.asset_metadata import FieldMetadata as ORMFieldMetadata
 from ..models.orm.assets import Asset as ORMAsset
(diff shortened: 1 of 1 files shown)


app/tasks/dynamic_vector_tile_cache_assets.py
```diff
@@ -7,6 +7,7 @@
 from ..models.enum.assets import AssetStatus, is_database_asset
 from ..models.enum.change_log import ChangeLogStatus
 from ..models.enum.creation_options import IndexType
+from ..models.orm.asset_metadata import FieldMetadata
```
danscales, line 10: In the other place where this is imported, it is done as:

```
from ..models.orm.asset_metadata import FieldMetadata as ORMFieldMetadata
```

so you might want to change to that (and change to ORMFieldMetadata below) for consistency.

danscales (approved):
Just one minor comment, but otherwise looks good to me, especially since it compiles and tests correctly with the new type hints.

app/tasks/dynamic_vector_tile_cache_assets.py
dmannarino: Thanks Dan! Although I think it has generally been done as a necessity when things conflict, you're right that there's value in consistency. I'll change it!

PR #421, opened 2023-09-05
GTC-2185 Handle area density rasters ending in ha_yr-1 

danscales:
GTC-2185 Handle area density rasters ending in ha_yr-1

We do special handling for rasters with pixel_meaning ending in "_ha-1",
do the same for rasters with pixel_meaning  ending in "_ha_yr-1".

I didn't add an extra test, because it looks like the one "_ha-1" test is
disabled because of issues with the caching of _get_data_environment.


dmannarino:
At the risk of over-complicating things, I'd be tempted to make AREA_DENSITY_RASTER_SUFFIXES a list or tuple and seeing if the end of the pixel meanings matches any of them. I just suspect that there may be more than these two in the future. So like
AREA_DENSITY_RASTER_SUFFIX1 = ("_ha-1", "_ha_yr-1")
...
for suffix in AREA_DENSITY_RASTER_SUFFIXES:
    if default_type.endswith(suffix):
        return f"{dataset}__{default_type[:len(suffix)]}"

What do you think?

dmannarino (approved):

gtempus (approved):
This is a big improvement! Good suggestion, @dmannarino. 💪 

*This is a giant file and we need to break it up (@dmannarino and I are working on this in another PR)*

PR #425, opened 2023-09-14
GTC-2185 Fix typos which meant _ha_yr-1 was still not handled

danscales:
GTC-2185 Fix typos which meant _ha_yr-1 was still not handled

Main thing to fix was a python indentation error, which meant that code was only checking the first suffix. But should also have a negative on the slice operation to remove the suffix.

Ran the '_ha-1' test by hand.

### DIFF
--- a/app/routes/datasets/queries.py
+++ b/app/routes/datasets/queries.py
@@ -83,7 +83,7 @@ router = APIRouter()
 
 
 # Special suffixes to do an extra area density calculation on the raster data set.
-AREA_DENSITY_RASTER_SUFFIXES = ["_ha-1", "ha_yr-1"]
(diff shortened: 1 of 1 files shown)


app/routes/datasets/queries.py
```diff
@@ -646,8 +646,8 @@ def _get_area_density_name(nm):
     return nm with the area-density suffix removed."""
     for suffix in AREA_DENSITY_RASTER_SUFFIXES:
         if nm.endswith(suffix):
```
jterry64, line 648: What nm mean again?

jterry64 (approved):

app/routes/datasets/queries.py
danscales: nm is the original pixel_meaning or source_layer_name before the possible magic suffix is removed.

PR #428, opened 2023-09-21
GTC-2443 Update to fixing /fields query

danscales:
GTC-2443 Update to fixing /fields query

The values_table in RasterBandMetadata should be specified as a dictionary to satisfy pydantic.

Tested by creating a sample RasterBandMetadata object by hand from a field_kwargs argument with some "value_tables" values (which we couldn't otherwise test).

### DIFF
--- a/app/routes/datasets/versions.py
+++ b/app/routes/datasets/versions.py
@@ -70,6 +70,7 @@ from ...tasks.delete_assets import delete_all_assets
 from ...utils.aws import get_aws_files
 from ...utils.google import get_gs_files
 from .queries import _get_data_environment
+from typing import cast
 
 router = APIRouter()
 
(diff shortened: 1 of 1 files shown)


solomon-negusse (approved):

app/routes/datasets/versions.py
```diff
@@ -479,11 +480,12 @@ async def _get_raster_fields(asset: ORMAsset) -> List[RasterBandMetadata]:
             "pixel_meaning": layer.name
         }
         if layer.raster_table:
-            field_kwargs["values_table"] = [
-                 row.meaning for row in layer.raster_table.rows
+            field_kwargs["values_table"] = cast(Dict[str, Any], {})
```
dmannarino, line 483: I don't think is quite the intended use of cast, which is generally supposed to inform type checkers like mypy and otherwise be a no-op. I'd really rather you just set it to an empty dictionary. If you want to add some type safety, may I suggest this (untested) overhaul of the function?

async def _get_raster_fields(asset: ORMAsset) -> List[RasterBandMetadata]:
    fields: List[RasterBandMetadata] = []

    # Magic field which is the area of the region satisfying the query
    fields.append(RasterBandMetadata(pixel_meaning="area__ha"))

    # Fetch all raster tile sets that have the same grid
    grid: Grid = Grid(asset.creation_options["grid"])
    raster_data_environment: DataEnvironment = await _get_data_environment(grid)

    logger.debug(f"Processing data environment {raster_data_environment}")
    for layer in raster_data_environment.layers:
        raster_band_metadata: RasterBandMetadata = RasterBandMetadata(
            pixel_meaning=layer.name,
            values_table=layer.raster_table
        )
        fields.append(raster_band_metadata)

    return fields

danscales: Yes, your version is much cleaner, and from the by-hand checking I can do, seems likely to work.  I will do that in a cleanup checkin later (since I merged just before your comments).  My current version seems to work fine (e.g. see output from https://staging-data-api.globalforestwatch.org/dataset/gfw_integrated_alerts/v20230824/fields , which was broken before).

PR #433, opened 2023-09-29
GTC-2161 Document special fields for raster-only datasets

danscales:
GTC-2161 Document special fields for raster-only datasets

Improve the documentation for query of raster-only datasets. In particular, mention that the special fields 'area__ha', 'latitude', and 'longitude' can be used and that a geometry must be specified for raster-only datasets.

Also, now that we know that 'latitude' and 'longitude' can be used for all raster-only datasets, add them to the actual /fields query (along with 'area__ha', which I added in a previous commit).

### DIFF
--- a/app/routes/datasets/versions.py
+++ b/app/routes/datasets/versions.py
@@ -467,8 +467,9 @@ async def _get_raster_fields(asset: ORMAsset) -> List[RasterBandMetadata]:
     fields: List[RasterBandMetadata] = []
 
     # Magic field which is the area of the region satisfying the query
-    args: Dict[str, Any] = {"pixel_meaning": "area__ha"}
-    fields.append(RasterBandMetadata(**args))
+    for magic in ["area__ha", "latitude", "longitude"]:
+        args: Dict[str, Any] = {"pixel_meaning": magic}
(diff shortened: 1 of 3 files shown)


solomon-negusse (approved):

app/routes/datasets/versions.py
```diff
@@ -467,8 +467,9 @@ async def _get_raster_fields(asset: ORMAsset) -> List[RasterBandMetadata]:
     fields: List[RasterBandMetadata] = []
 
     # Magic field which is the area of the region satisfying the query
-    args: Dict[str, Any] = {"pixel_meaning": "area__ha"}
-    fields.append(RasterBandMetadata(**args))
+    for magic in ["area__ha", "latitude", "longitude"]:
```
jterry64, line 470: I think I missed the original PR. While I kind of love calling it magic, I feel like something like "reserved_fields" would be more descriptive. But honestly this also makes sense, so whatever you think works best

jterry64 (approved):

app/routes/datasets/versions.py
danscales: Thanks for the comment!  Yes, changed "magic" to "reserved" in the code and in the documentation.

PR #436, opened 2023-10-10
GTC-2576 Restrict queries on licensed WDPA dataset

danscales:
GTC-2576 Restrict queries on licensed WDPA dataset

Only allow queries to versions of the licensed WDPA dataset for admin users (i.e. a bearer token is provided that signifies an admin user).

For now, we just have a simple list of restricted datasets in the code, rather than adding some new dataset attribute.

Added a test licensed dataset/version and added a test that the query is restricted of a licensed version.

### DIFF
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -28,41 +28,37 @@ async def is_service_account(token: str = Depends(oauth2_scheme)) -> bool:
         return True
 
 
-# Check is the authorized user is an admin. Return true if so, throw
-# an exception if not.
 async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
     """Calls GFW API to authorize user.
 
     User must be ADMIN for gfw app
     """
 
-    response: Response = await who_am_i(token)
+    return await is_app_admin(token, "gfw", "Unauthorized")
 
-    if response.status_code == 401 or not (
-        response.json()["role"] == "ADMIN"
-        and "gfw" in response.json()["extraUserData"]["apps"]
-    ):
-        logger.warning(f"ADMIN privileges required. Unauthorized user: {response.text}")
-        raise HTTPException(status_code=401, detail="Unauthorized")
-    else:
-        return True
-
-# Check is the authorized user is an admin. Return true if so, false if not (with no
-# exception).
-async def is_admin_no_exception(token: str = Depends(oauth2_scheme)) -> bool:
+async def is_gfwpro_admin(error_str: str, token: str = Depends(oauth2_scheme)) -> bool:
     """Calls GFW API to authorize user.
 
-    User must be ADMIN for gfw app
+    User must be ADMIN for gfw pro app
+    """
+
+    return await is_app_admin(token, "gfw-pro", error_str)
+
+async def is_app_admin(token: str, app: str, error_str: str) -> bool:
+    """Calls GFW API to authorize user.
+
+    User must be an ADMIN for the specified app, else it will throw
+    an exception with the specified error string.
     """
 
     response: Response = await who_am_i(token)
 
     if response.status_code == 401 or not (
         response.json()["role"] == "ADMIN"
-        and "gfw" in response.json()["extraUserData"]["apps"]
+        and app in response.json()["extraUserData"]["apps"]
     ):
         logger.warning(f"ADMIN privileges required. Unauthorized user: {response.text}")
-        return False
+        raise HTTPException(status_code=401, detail=error_str)
     else:
         return True
 
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -28,6 +28,8 @@ async def is_service_account(token: str = Depends(oauth2_scheme)) -> bool:
         return True
 
 
+# Check is the authorized user is an admin. Return true if so, throw
+# an exception if not.
 async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
     """Calls GFW API to authorize user.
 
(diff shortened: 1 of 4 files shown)


app/authentication/token.py
```diff
@@ -45,6 +47,25 @@ async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
     else:
         return True
 
+# Check is the authorized user is an admin. Return true if so, false if not (with no
+# exception).
+async def is_admin_no_exception(token: str = Depends(oauth2_scheme)) -> bool:
```
solomon-negusse, line 52: given how similar this is to `is_admin` above, should we just add `no_exception=bool` keyword argument to `is_admin`

app/authentication/token.py
```diff
@@ -45,6 +47,25 @@ async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
...
+    """Calls GFW API to authorize user.
+
+    User must be ADMIN for gfw app
+    """
+
+    response: Response = await who_am_i(token)
+
+    if response.status_code == 401 or not (
+        response.json()["role"] == "ADMIN"
+        and "gfw" in response.json()["extraUserData"]["apps"]
```
solomon-negusse, line 62: if this dataset is to be used just by GFW Pro, would could consider further narrowing down access to users with `gfw-pro` in their `apps` object.

solomon-negusse:
@danscales, left a couple of comments. Also, not part of your change, but I noticed `_check_downloadability` function [checks](https://github.com/wri/gfw-data-api/blob/bd497d5509203b0dc353b5325fd6a35d72f10556/app/routes/datasets/downloads.py#L335) for `is_downloadable` attribute on the version record which doesn't have that attribute. May be worth checking it behaves as expected.

danscales:
> @danscales, left a couple of comments. Also, not part of your change, but I noticed `_check_downloadability` function [checks](https://github.com/wri/gfw-data-api/blob/bd497d5509203b0dc353b5325fd6a35d72f10556/app/routes/datasets/downloads.py#L335) for `is_downloadable` attribute on the version record which doesn't have that attribute. May be worth checking it behaves as expected.

Thanks for your review.  I'll talk to you in person about your suggestion on checking for "gfwpro" in the apps object.

With respect to _check_downloadability(), as far as I can tell, Version does have a field 'is_downloadable' which is populated (see version.py, line 10) and propagated from the dataset 'is_downloadable' field if necessary (see create_version in version.py).

Also, https://data-api.globalforestwatch.org/dataset/wdpa_protected_areas/v202308/download/json?sql=%22SELECT(*)%20from%20mytable%22  already correctly returns a error saying the dataset is not downloadable.

app/routes/datasets/queries.py
```diff
@@ -85,6 +86,10 @@
 # Special suffixes to do an extra area density calculation on the raster data set.
 AREA_DENSITY_RASTER_SUFFIXES = ["_ha-1", "_ha_yr-1"]
 
+# Datasets that require admin privileges to do a query. (Extra protection on
+# commercial datasets which shouldn't be downloaded in any way.)
+PROTECTED_QUERY_DATASETS = ["licensed_wdpa_protected_areas"]
```
jterry64, line 91: I'm not sure how I feel about hardcoding this. If we need this very soon I wouldn't temporarily, but long term, I feel like we should make a general options to restrict access. 

@manukala6 curious, are there any things not downloadable that it's ok to query? Maybe that flag should just restrict viewing anything but metadata.

app/routes/datasets/queries.py
```diff
@@ -155,6 +160,9 @@ async def query_dataset_json(
     """
 
     dataset, version = dataset_version
+    if dataset in PROTECTED_QUERY_DATASETS:
+        await is_gfwpro_admin(error_str="Unauthorized query on a restricted dataset")
```
jterry64, line 164: Similar to above comment, do we really want to get in the business of restricting hardcoded datasets to hardcoded users? Seems like it goes against the general design of the API, and we'll need to redeploy every time we need to add new datasets or users to these lists.

Long term solution seems more like restricted datasets should only be accessible by the admin user that uploaded it.

But once again, that's pretty complicated, and this situation has only come up once. So ok with this as a stopgap until we start seeing this more regularly from Pro or other users.

jterry64 (changes requested):

app/authentication/token.py
danscales: Yes, I separated out a function is_app_admin() that handles case checking for gfw or gfwpro admin.  I didn't need the no_exception part (that was really for an earlier version), since I really do want an exception, as long as I can customize the message.

app/authentication/token.py
danscales: Yes, I changed for checking for the gfw-pro admin in the case of the licensed WDPA dataset.  Thanks for the suggestions!

app/routes/datasets/queries.py
danscales: It sounds from both of your comments that you are OK with doing this for now, so I'll go ahead with this check-in.  I had discussed with other folks, and they were OK as well, given that this is just a single dataset and we're not sure if we will need to protect any other datasets in this way.  A more general solution would require adding a whole new 'restricted_query' attribute to the data API DB schema.

We talked about restricting queries for all versions marked with is_downloadable as false.  However, there are already 8 different datasets with is_downloadable as false, including the public WDPA dataset.  So, I think we would definitely be affecting some current users if we restricted queries to all datasets versions with is_downloadable as FALSE.

PR #454, opened 2023-12-29
GTC-2618 Allow vector appends from new layers

manukala6:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Appending polygons to a vector source assumes the original creation options, meaning it's not possible to append selected layers from a new file

Issue Number: [GTC-2618](https://gfw.atlassian.net/browse/GTC-2618)

## What is the new behavior?

- Allows append of specific input layers to vector dataset
- Allows multiple source URIs for GPKGs, FileGDBs in creation options

- [ ] Yes
- [x] No

## Other information

[GTC-2618]: https://gfw.atlassian.net/browse/GTC-2618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmannarino:
Hmm, will this section of code need some modification? https://github.com/wri/gfw-data-api/blob/master/app/tasks/vector_source_assets.py#L361-L386
I think it may try to add the original layers again? Not sure. Could you make a test to verify?

app/routes/datasets/versions.py
```diff
@@ -232,13 +231,18 @@ async def append_to_version(
     # For the background task, we only need the new source uri from the request
     input_data = {"creation_options": deepcopy(default_asset.creation_options)}
     input_data["creation_options"]["source_uri"] = request.source_uri
+    # check if layers are provided and append to creation options
+    if input_data["creation_options"].get("layers"):
+        input_data["creation_options"]["layers"] = request.layers
```
danscales, line 236: Shouldn't the `if` be testing the value of request.layers (to be non-None), rather than the value of layers that came from the default asset?

With the current code, if the existing default asset has some layers, and request.layers is None (no layers specified), then you will set input_data["creation_options"]["layers"] to None, but I think you want to keep it at the current layers, right?  (Hopefully I'm understanding the semantics of pydantic fields correctly.)

app/routes/datasets/versions.py
```diff
@@ -232,13 +231,18 @@ async def append_to_version(
...
     background_tasks.add_task(
         append_default_asset, dataset, version, input_data, default_asset.asset_id
     )
 
     # We now want to append the new uris to the existing ones and update the asset
     update_data = {"creation_options": deepcopy(default_asset.creation_options)}
     update_data["creation_options"]["source_uri"] += request.source_uri
+    if input_data["creation_options"].get("layers"):
+        update_data["creation_options"]["layers"] += request.layers
```
danscales, line 245: Also, if request.layers is None, seems weird to append None to the layers list here, so maybe you want to also check in the `if` statement if request.layers is non-None?  Again, I might be mistunderstanding pydantic semantics here.

jterry64: Yeah, I think Dan is correct. Likewise, what happens if the user is appending a different source URI with a different layer name? Is it just listing all source URIs and layers? That's mostly fine but might be a little confusing to read.

app/models/pydantic/creation_options.py
```diff
@@ -225,10 +225,10 @@ class VectorSourceCreationOptions(StrictBaseModel):
     def validate_source_uri(cls, v, values, **kwargs):
         if values.get("source_driver") == VectorDrivers.csv:
             assert len(v) >= 1, "CSV sources require at least one input file"
-        else:
+        elif values.get("source_driver") in [VectorDrivers.esrijson, VectorDrivers.shp, VectorDrivers.geojson_seq, VectorDrivers.geojson]:
```
jterry64, line 228: This is a little tricky, because I don't think we actually support multiple geodatabase/filegdbs, we just support appending a different file. But if a user actually sent in multiple filegdbs with layers from each, I think it'd break. We have no way to know which layers are from which files.

So I think we do still want to enforce one source URI per operation? (new version or append)

tests/routes/datasets/test_versions.py
```diff
@@ -348,6 +349,19 @@ async def test_invalid_source_uri(async_client: AsyncClient):
         == f"Version with name {dataset}.{bad_version} does not exist"
     )
 
+    # Test appending a layer to a version
+    response = await async_client.post(
+        f"/dataset/{dataset}/{version}/append", json={"source_uri": f"s3://{BUCKET}/{GPKG_NAME}", "layers": ["layer2"]}
+    )
+    assert response.status_code == 200
+
```
jterry64, line 357: so we only support appending if the original version had layers specified?

jterry64 (changes requested):

manukala6:
Changes moved to feature/vector_layer_appends. Closing this PR

PR #456, opened 2024-01-02
Validate src_uris concurrently

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


app/routes/datasets/versions.py
```diff
@@ -528,7 +525,11 @@ def _verify_source_file_access(sources: List[str]) -> None:
 
     for source in sources:
         url_parts = urlparse(source, allow_fragments=False)
-        list_func = source_uri_lister_constructor[url_parts.scheme.lower()]
+        try:
```
jterry64, line 528: As mentioned in our huddle, I think it'd solve the problem better to make the invalid sources check a background task that just sets the version to failed rather than doing it upfront. But also might require better error messaging in the version metadata. 

Alternative workaround: have datapump just send a directory rather than all the source URIs.

jterry64 (changes requested):

PR #469, opened 2024-02-14
GTC-2706 Give error code and message when asset UID is wrong

danscales:
GTC-2706 Give error code and message when asset UID is wrong

If the asset uid when creating a raster tile cache is wrong (bad format or nonexistent), return a better status code and error message, rather than just a 500 (internal system error)

Added new test in test_assets.py for error cases when the source asset uid does not exist. I factored out a new function for creating the default asset to reduce a bunch of duplicated code.

Also, the test test_raster_tile_cache_asset had been completely disabled for quite a while. I re-enabled the first 4 sub-tests, which work fine. The fifth sub-test does not work, so I left it disabled and filed issue GTC-2735.

### DIFF
--- a/app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
+++ b/app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
@@ -4,6 +4,7 @@ from uuid import UUID
 import numpy as np
 from fastapi import HTTPException
 from fastapi.logger import logger
+from asyncpg import DataError
 
 from app.crud.assets import get_asset
 from app.models.enum.assets import AssetType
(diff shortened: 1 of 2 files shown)


jterry64 (approved):

app/tasks/raster_tile_cache_assets/raster_tile_cache_assets.py
```diff
@@ -204,9 +206,15 @@ async def raster_tile_cache_validator(
...
-    )
+    try:
+        source_asset: ORMAsset = await get_asset(
+            input_data["creation_options"]["source_asset_id"]
+        )
+    except RecordNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e))
+    except DataError as e:
+        raise HTTPException(status_code=400, detail=str(e))
+
```
dmannarino, line 217: What causes a DataError? Should we log it too?

danscales: DataError is if the asset uuid isn't even the right format (e.g. too short).  The detail does have a correct useful message, and I'm returning a 400 error code because it's a client input mistake.

dmannarino (approved):
Great, thanks for fixing this and doing the tests work!

PR #478, opened 2024-02-28
Resuscitate old branch to delete its resources (and improve GHA triggers)

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

- A shadow production deploy exists, an artifact of overly liberal GHA workflows

Issue Number: GTC-2152

## What is the new behavior?

- Slightly improve the GHA workflows, but most importantly when this branch is merged and deleted it should take down the shadow deploy.

- [ ] Yes
- [x] No


.github/workflows/terraform_plan.yaml
```diff
@@ -9,15 +9,16 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Plan production
-      if: success() && contains(github.base_ref, 'master')
```
solomon-negusse, line 12: may have to double-check this. `base_ref` of a pull request trigger will point to the target branch where we run the plan but `ref_name` will point to the "feature branch".

solomon-negusse: Ah i misread, i see you're using `github.ref`. Still would be good to clarify why this change was needed since `ref` looks less straightforward than`base_ref`

dmannarino: Thank you for catching that! Changed to use == base_ref

solomon-negusse (approved):

PR #488, opened 2024-03-27
GTC-2449 Catch access denied error in get_aws_files()

danscales:
GTC-2449 Catch access denied error in get_aws_files()

Currently, we are only catching NoSuchBucket errors, but there can also be AccessDenied errors. If an 'access denied' happens, we get an uncaught exception, which gives an internal server error. Strangely, s3_client.exceptions doesn't have AccessDenied, so we just catch all botocore exceptions (which is probably what we should do anyway).

I couldn't create any new test to emulate access denied, so will just test it out in staging.

### DIFF
--- a/app/routes/datasets/versions.py
+++ b/app/routes/datasets/versions.py
@@ -560,7 +560,7 @@ def _verify_source_file_access(sources: List[str]) -> None:
         raise HTTPException(
             status_code=400,
             detail=(
-                "Cannot access all of the source files. "
+                "Cannot access all of the source files (non-existent or access denied). "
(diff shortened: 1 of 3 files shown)


danscales:
FYI, the tests all succeeded, but the test failure is from the usual problem with a limit of 50 job queues in the dev test environment.

jterry64 (approved):

dmannarino (approved):
Cool, thanks for fixing this!

solomon-negusse (approved):

danscales:
I forgot to change the destination branch from master to develop, so closing this PR and opening another one where the destination branch is develop.

PR #489, opened 2024-03-27
Allowing updating raster metadata

jterry64:
## Pull request checklist

## Pull request type

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, you can make include raster asset metadata on asset creation through a  `PATCH` request. However, you can't update existing raster asset metadata or we return a `Internal Server Error`. Looks like this code path wasn't updated when we did the big metadata update.

Pushing as a hotfix because I accidentally put the wrong raster metadata for TCL 2023 to work with OTF (forgot to add the year 2023 to the value table). I'd prefer to just fix it now and update it through the API than having to do some hacky workaround.

## What is the new behavior?
Now you can update existing raster asset metadata. However, one caveat is that we allow metadata for each band, but the bands are just a list. There's no ID, so there's no way to update only one band as far as I can tell. So you need to at least include all bands in the update request in order. Open to people's thoughts on this.

- [ ] Yes
- [X] No


dmannarino (approved):
Thanks for fixing this, and especially the test!

solomon-negusse:
@jterry64 there is an orm [model](https://github.com/wri/gfw-data-api/blob/master/app/models/orm/asset_metadata.py#L45) for raster bands, just didn't exposed for CRUD :/. So we should update the raster bands individually the same way  we do for `fields` (https://github.com/wri/gfw-data-api/blob/2122dfdbe8a5f5df23022e098b0b4b5aacec0b4a/app/crud/metadata.py#L202). This won't need passing all the bands to update/add just one and will handle duplicates, etc.

solomon-negusse (changes requested):
@jterry64 there is an orm [model](https://github.com/wri/gfw-data-api/blob/master/app/models/orm/asset_metadata.py#L45) for raster bands, just didn't exposed for CRUD :/. So we should update the raster bands individually the same way  we do for `fields` (https://github.com/wri/gfw-data-api/blob/2122dfdbe8a5f5df23022e098b0b4b5aacec0b4a/app/crud/metadata.py#L202). This won't need passing all the bands to update/add just one and will handle duplicates, etc.

jterry64:
> @jterry64 there is an orm [model](https://github.com/wri/gfw-data-api/blob/master/app/models/orm/asset_metadata.py#L45) for raster bands, just didn't exposed for CRUD :/. So we should update the raster bands individually the same way we do for `fields` (
> 
> https://github.com/wri/gfw-data-api/blob/2122dfdbe8a5f5df23022e098b0b4b5aacec0b4a/app/crud/metadata.py#L202
> 
> ). This won't need passing all the bands to update/add just one and will handle duplicates, etc.

But how will users be able to specify which band their updating in their `PATCH` request?

solomon-negusse:
>But how will users be able to specify which band their updating in their PATCH request?
pixel_meaning is the primary key

jterry64:
Confirmed it works in staging without a migration, so going to merge

PR #496, opened 2024-04-15
Cache _get_data_env results for 5 minutes

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- _get_data_environment (which is called every time a query is made) performs a (slow) DB query to look for assets, despite the fact that the information does not change often.

Issue Number: GTC-2779

## What is the new behavior?

- The results of _get_data_environment are cached with a time to live of 5 minutes. This should eliminate a significant portion of queries to the DB without causing much lag in acknowledging new assets.

- [ ] Yes
- [x] No

### DIFF
--- a/tests_v2/unit/app/routes/analysis/test_analysis.py
+++ b/tests_v2/unit/app/routes/analysis/test_analysis.py
@@ -79,7 +79,7 @@ async def test_analysis_with_huge_geostore(
 
 @pytest.mark.asyncio
 async def test_raster_analysis_payload_shape(
-    generic_dataset, async_client_per_function: AsyncClient, monkeypatch: MonkeyPatch
+    generic_dataset, async_client: AsyncClient, monkeypatch: MonkeyPatch
 ):
     """Note that this test depends on the output of _get_data_environment
     which will likely have cached values from other tests, so we clear it."""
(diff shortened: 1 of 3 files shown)


solomon-negusse (approved):
Awesome! Just one minor suggestion (maybe for future iteration): what if we clear the cache every time we add a new raster tile set and set `ttl` to a value much larger the 5 mins you set.

dmannarino:
> > Awesome! Just one minor suggestion (maybe for future iteration): what if we clear the cache every time we add a new raster tile set (or set one to latest) and set `ttl` to a value much larger the 5 mins you set.
> 
> Oh, man, that's a great idea. Let me see if I can sneak it in quickly. :)
Actually, I won't right now. I need to think about how to do it right rather than making spaghetti code.
An inclination supported by the fact that my attempt resulted in a circular dependency. ;) So I'll create follow-up ticket and merge this as is. Thanks!

jterry64 (approved):
Sweet, glad it was just basically a one-liner

tests_v2/unit/app/routes/analysis/test_analysis.py
```diff
@@ -103,6 +106,9 @@ async def test_raster_analysis_payload_shape(
         )
         monkeypatch.setattr(geostore.rw_api, "get_geostore", mock_rw_get_geostore)
 
+        # The other tests will have polluted the data env cache. Clear it.
+        queries._get_data_environment.cache_clear()
```
gtempus, line 110: Cool! I learned something new. Nice work, @dmannarino!

PR #504, opened 2024-05-02
GTC-2027 (part 2) - give useful error when input geometry is not Polygon/MultiPolygon

danscales:
GTC-2027 (part 2) - give useful error when input geometry is not Polygon/MultiPolygon

This is to fix for a bunch more "tuple index out of range" 500 errors that are showing up.  Someone is submitting zonal statistic queries on geometries which are not Polygons.  This is the same as my previous fix for normal queries.

Since zonal statististics are on the way out (eventually), I didn't add a new test for this particular error case.

### DIFF
--- a/app/routes/analysis/analysis.py
+++ b/app/routes/analysis/analysis.py
@@ -101,6 +101,12 @@ async def _zonal_statistics(
     start_date: Optional[str],
     end_date: Optional[str],
 ):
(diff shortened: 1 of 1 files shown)


jterry64 (approved):

danscales:
Thanks for the review!

PR #505, opened 2024-05-02
Allow admins to reassign dataset owner

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Dataset owners are assigned on dataset creation but can't be changed through the API.

Issue Number: GTC-2800

## What is the new behavior?
Admin users can now send update the owner_id PATCH requests to a dataset.

- [ ] Yes
- [X] No

### DIFF
--- a/app/routes/datasets/dataset.py
+++ b/app/routes/datasets/dataset.py
@@ -56,11 +56,12 @@ async def create_dataset(
     owner_id: str = Depends(rw_user_id),
     response: Response,
 ) -> DatasetResponse:
-    """Create a dataset. A “dataset” is largely a metadata concept: it represents
-    a data product that may have multiple versions or file formats over time.
-    The user that creates a dataset using this operation becomes the owner of
-    the dataset, which provides the user with the privileges to do further write
-    operations on the dataset, including creating and modifying versions and assets.
+    """Create a dataset. A “dataset” is largely a metadata concept: it
+    represents a data product that may have multiple versions or file formats
+    over time. The user that creates a dataset using this operation becomes the
+    owner of the dataset, which provides the user with the privileges to do
+    further write operations on the dataset, including creating and modifying
+    versions and assets.
 
     This operation requires a `MANAGER` or an `ADMIN` user role.
     """
@@ -94,15 +95,24 @@ async def update_dataset(
     dataset: str = Depends(dataset_dependency),
     request: DatasetUpdateIn,
     is_authorized: bool = Depends(is_admin),
+    is_authorized_admin: bool = Depends(is_admin),
 ) -> DatasetResponse:
     """Partially update a dataset.
 
     Only metadata field can be updated. All other fields will be
     ignored.
 
-    Only the dataset owner or a user with `ADMIN` user role can do this operation.
+    Only the dataset owner or a user with `ADMIN` user role can do this operation. Only
+    a user with `ADMIN` role can change the dataset owner.
     """
     input_data: Dict = request.dict(exclude_none=True, by_alias=True)
+
+    if request.owner_id is not None and not is_authorized_admin:
+        raise HTTPException(
+            status_code=401,
+            detail="Only ADMIN users can change owner_id. Please contact an ADMIN if you would like change dataset ownership.",
+        )
+
     row: ORMDataset = await datasets.update_dataset(dataset, **input_data)
 
     return await _dataset_response(dataset, row)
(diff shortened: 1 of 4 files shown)


gtempus (changes requested):

gtempus:
I'm so upset at GitHub right now. I had more info to share in the PR review general comment roll-up and it just deleted it. 😡 
I'll add those thoughts tomorrow.

app/routes/datasets/dataset.py
```diff
@@ -94,15 +95,24 @@ async def update_dataset(
     dataset: str = Depends(dataset_dependency),
     request: DatasetUpdateIn,
     is_authorized: bool = Depends(is_admin),
+    is_authorized_admin: bool = Depends(is_admin),
```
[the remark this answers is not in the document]
jterry64: @gtempus yes, this was anticipating Daniel's change, which would change the above line to be checking for admin or manager role. I think I'll wait for his PR which will separate `is_admin` and `assert_admin` like we discussed so this does actually start returning true/false.

app/routes/datasets/dataset.py
```diff
@@ -94,15 +95,24 @@ async def update_dataset(
...
 ) -> DatasetResponse:
     """Partially update a dataset.
 
     Only metadata field can be updated. All other fields will be
     ignored.
```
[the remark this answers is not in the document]
jterry64: good catch, I think this will already incorrect for a while because of `is_downloadable`

app/routes/datasets/dataset.py
```diff
@@ -94,15 +95,24 @@ async def update_dataset(
...
 
-    Only the dataset owner or a user with `ADMIN` user role can do this operation.
+    Only the dataset owner or a user with `ADMIN` user role can do this operation. Only
+    a user with `ADMIN` role can change the dataset owner.
     """
     input_data: Dict = request.dict(exclude_none=True, by_alias=True)
+
+    if request.owner_id is not None and not is_authorized_admin:
```
solomon-negusse, line 110: is there a check here to see if the new `owner_id` has a `MANAGER` role or are we missing that?

jterry64: good caall

app/utils/rw_api.py
```diff
@@ -84,6 +84,37 @@ async def who_am_i(token) -> Response:
...
+    except ReadTimeout:
+        raise HTTPException(
+            status_code=500,
+            detail="Call to user service timed-out. Please try again.",
+        )
+
+    if response.status_code == 404:
+        raise HTTPException(status_code=401, detail=f"User ID invalid: {user_id}")
+
+    if response.status_code != 200 and response.status_code != 401:
```
danscales, line 107: I'm not sure why you are explicitly exempting status_code 401 here.  It seems like that is a real response from /auth/user/{user_id}: https://resource-watch.github.io/doc-api/reference.html#get-the-current-user.  We might get that error if our service account has expired?  In any case, seems like you should raise an exception for any status code other than 200, since response won't be correct otherwise.

tests/routes/datasets/test_datasets.py
```diff
@@ -70,16 +73,54 @@ async def test_datasets(async_client):
...
+            )
+
+        assert len(rows) == 1
+        assert rows[0][0] == "new_owner_id123"
+
+        # assign original owner back
+        from app.main import app
+
+        app.dependency_overrides[get_manager] = get_new_owner_mocked
+        new_payload = {"owner_id": "mr_manager123"}
```
danscales, line 102: To be fully correct, I would think that you need to put this code from line 101 to 104 inside a different 'with patch' that forces get_rw_user to return MANAGER (i.e. the User object that get_manager_mocked is returning), so it matches up with the owner_id "mr_manager123" that you are going to be looking up.  The test works now because NEW_OWNER is also a manager, which is what you require.

Feel free to ignore this for now.

danscales (approved):
Looks good, but I do have a comment on the tests and still the question about why you are explicitly ignoring status code 401 in get_rw_user().

app/routes/datasets/dataset.py
```diff
@@ -21,21 +21,23 @@
...
 router = APIRouter()
 
 
 async def get_owner(
     dataset: str = Depends(dataset_dependency), user: User = Depends(get_manager)
 ) -> User:
-    """Returns the User making the request as long as that user is an admin or
-    the owner of the dataset, otherwise raises a 401."""
+    """Retrieves the user object that owns the dataset, or ADMIN user, if that
+    user is the one making the request, otherwise raises a 401."""
```
dmannarino, line 33: This is a little clunky sounding. Perhaps we could re-word as "Retrieves the user object of the one making the request if that user either owns the dataset or is an ADMIN, otherwise raises a 401"
What do you think?

PR #506, opened 2024-05-03
WIP: Mutate fcns require owner or admin

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -52,6 +52,7 @@ async def is_admin(token: str = get_request_token()) -> bool:
 
     return await is_app_admin(token, "gfw", "Unauthorized")
 
+
 async def is_gfwpro_admin_for_query(dataset: str = Depends(dataset_dependency),
                                     token: str | None = get_request_token_optional()) -> bool:
     """If the dataset is protected dataset, calls GFW API to authorize user by
@@ -67,6 +68,7 @@ async def is_gfwpro_admin_for_query(dataset: str = Depends(dataset_dependency),
                                error_str="Unauthorized query on a restricted dataset")
 
     return True
+
 
 async def is_app_admin(token: str, app: str, error_str: str) -> bool:
     """Calls GFW API to authorize user.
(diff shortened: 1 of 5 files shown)


app/authentication/token.py
```diff
@@ -91,3 +116,27 @@ async def get_user(token: str = Depends(oauth2_scheme)) -> Tuple[str, str]:
         raise HTTPException(status_code=401, detail="Unauthorized")
     else:
         return response.json()["id"], response.json()["role"]
+
+
+async def user_is_owner_or_admin(dataset: str) -> bool:
+    user_id, role = await get_request_user()
+
+    if role == "ADMIN":
```
jterry64, line 124: Not sure if we still want to go for this, but technically we check if ADMIN is also "gfw" application. Did we still want to check an application, or should RW ADMINs just have universal access?

app/authentication/token.py
```diff
@@ -4,17 +4,31 @@
...
+from ..crud import datasets
+from ..models.orm.datasets import Dataset as ORMDataset
+from ..routes import dataset_dependency
 from ..settings.globals import PROTECTED_QUERY_DATASETS
+from ..utils.rw_api import who_am_i
+
+
+def get_request_token() -> str:
+    """Get request token if it exists."""
+    return str(OAuth2PasswordBearer(tokenUrl="/token"))
```
danscales, line 17: I'm not sure this will now work the way you have restructured it.  The Oauth2PasswordBearer() object needs to be referenced as part of a Depends dependency to actually do its checking of the current request token.  As far as I know, you are just creating an instance of OAuth2PasswordBearer and then converting it to a string.  It's only when used as a dependency that it will do the checking via the __call_ method.  See https://github.com/tiangolo/fastapi/blob/e04d397e3224d3674cf73332303b956fc3d00b3d/fastapi/security/oauth2.py#L473 .  Also, you only need to create the oauth2_schema and oauth2_schema_no_auto objects once.  I would go back to using those pre-calculated objects.  I can huddle with you if that's not clear.

danscales (commented):
See my comment - I don't think your restructured auth checks are right.

PR #509, opened 2024-05-06
Pass user details when authenticating instead of a bool, check for manager on dataset creation

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently we just return a bool when checking authentication for most routes, except we really just throw an exception if the check fails.

Issue Number: GTC-2820

## What is the new behavior?

- Return a User object which contains all the details of a user when authenticating/authorizing, or throw an exception if the user isn't authorized. This is more in line with what FastAPI has in their [auth example](https://fastapi.tiangolo.com/tutorial/security/get-current-user/). This makes more sense than a boolean that never returns false, and provides an object with details we can use for other behavior (e.g. save ID as owner, return email of owner in error message, check orgs/apps, etc.)
- Add an initial use for setting the manager as owner when creating a dataset.

- [ ] Yes
- [X] No

### DIFF
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -33,13 +33,31 @@ async def is_service_account(token: str = Depends(oauth2_scheme)) -> bool:
         return True
 
 
-async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
+async def assert_admin(token: str = Depends(oauth2_scheme)) -> None:
     """Calls GFW API to authorize user.
 
     User must be ADMIN for gfw app
     """
 
-    return await is_app_admin(token, "gfw", "Unauthorized")
+    return await assert_app_role(token, "ADMIN", "gfw", "Unauthorized")
+
+
+async def assert_manager(token: str = Depends(oauth2_scheme)) -> None:
+    """Calls GFW API to authorize user.
+
+    User must be MANAGER for data-api app.
+    """
+
+    return await assert_app_role(token, "MANAGER", "data-api", "Unauthorized")
+
+
+async def is_admin_or_manager(token: str = Depends(oauth2_scheme)) -> bool:
+    """Calls GFW API to authorize user.
+
+    User must be ADMIN for gfw app or MANAGER for data-api app.
+    """
+
+    return (await assert_admin(token)) or (await assert_manager(token))
 
 
 async def rw_user_id(token: str = Depends(oauth2_scheme)) -> str:
@@ -65,8 +83,9 @@ async def is_gfwpro_admin_for_query(
                 status_code=401, detail="Unauthorized query on a restricted dataset"
             )
         else:
-            return await is_app_admin(
+            return await is_app_role(
                 cast(str, token),
+                "ADMIN",
                 "gfw-pro",
                 error_str="Unauthorized query on a restricted dataset",
             )
@@ -74,21 +93,23 @@ async def is_gfwpro_admin_for_query(
     return True
 
 
-async def is_app_admin(token: str, app: str, error_str: str) -> bool:
-    """Calls GFW API to authorize user.
+async def assert_app_role(token: str, role: str, app: str, error_str: str) -> None:
+    is_authorized = await is_app_role(token, role, app)
 
-    User must be an ADMIN for the specified app, else it will throw an
-    exception with the specified error string.
-    """
+    if not is_authorized:
+        raise HTTPException(status_code=401, detail=error_str)
+
+
+async def is_app_role(token: str, role: str, app: str) -> bool:
+    """Calls RW API to authorize user for specific role and app."""
 
     response: Response = await who_am_i(token)
 
     if response.status_code == 401 or not (
-        response.json()["role"] == "ADMIN"
+        response.json()["role"] == role
         and app in response.json()["extraUserData"]["apps"]
     ):
-        logger.warning(f"ADMIN privileges required. Unauthorized user: {response.text}")
-        raise HTTPException(status_code=401, detail=error_str)
+        return False
     else:
         return True
 
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -1,10 +1,11 @@
-from typing import Tuple, cast
+from typing import cast
 
 from fastapi import Depends, HTTPException
 from fastapi.logger import logger
 from fastapi.security import OAuth2PasswordBearer
 from httpx import Response
 
+from ..models.pydantic.authentication import User
 from ..routes import dataset_dependency
 from ..settings.globals import PROTECTED_QUERY_DATASETS
 from ..utils.rw_api import who_am_i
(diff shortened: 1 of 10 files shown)


app/authentication/token.py
```diff
@@ -93,19 +88,34 @@ async def is_app_admin(token: str, app: str, error_str: str) -> bool:
...
 
-    if response.status_code == 401 or not (
-        "gfw" in response.json()["extraUserData"]["apps"]
-    ):
+    if response.status_code == 401:
         logger.info("Unauthorized user")
         raise HTTPException(status_code=401, detail="Unauthorized")
     else:
-        return response.json()["id"], response.json()["role"]
+        return User(**response.json()["data"])
```
danscales, line 100: I may be missing something, but I'm not following this.  I don't see that you changed who_am_i(), so it is just returning the base level response from "/auth/check-logged".  But that doesn't have a "data" value.  It does have "email" and "extraUserData" and "role"  and "id", as shown by previous code in the file.  Won't the access to "data" just fail?

https://resource-watch.github.io/doc-api/reference.html#get-lt-base-api-url-gt-auth-check-logged

jterry64: @danscales  oh good catch, I copied it from the other line doing this, which calls `/auth/sign-up`. That returns it in a `data` object, but for some reason this endpoint doesn't. Which is weird, I thought that was just a standard (from `jsend`) across the entire RW API.  Fixed bow.

tests_v2/utils.py
```diff
@@ -32,15 +33,39 @@ def submit_batch_job(self, *args, **kwargs) -> uuid.UUID:
...
+        createdAt="2021-06-13T03:18:23.000Z",
+        role="ADMIN",
+        applications=[],
+        extraUserData={},
+    )
 
 
-async def get_rw_user_id_mocked() -> str:
-    return "userid_123"
+async def get_manager_mocked() -> str:
```
dmannarino, line 59: The tiniest thing, this should be -> User.

dmannarino (approved):
One type hint complaint, otherwise looks great!

app/authentication/token.py
```diff
@@ -93,19 +88,34 @@ async def is_app_admin(token: str, app: str, error_str: str) -> bool:
...
+
+    return user
+
+
+async def get_manager(user: User = Depends(get_user)) -> User:
+    """Get the details for authenticated MANAGER for data-api application or
+    ADMIN user."""
+
+    if user.role != "ADMIN" or not (
+        user.role == "MANAGER" and "data-api" in user.applications
```
danscales, line 117: How does user.applications get populated?  I see that the RW checked-logged returns "extraUserData.apps", not "applications".  https://resource-watch.github.io/doc-api/reference.html#get-lt-base-api-url-gt-auth-check-logged

[Oh, I see this again might be a leftover again from auth/sign-up, which does return an "applications" field, but auth/check-logged doesn't.]

app/authentication/token.py
```diff
@@ -93,19 +88,34 @@ async def is_app_admin(token: str, app: str, error_str: str) -> bool:
...
 
-    if response.status_code == 401 or not (
-        "gfw" in response.json()["extraUserData"]["apps"]
-    ):
+    if response.status_code == 401:
         logger.info("Unauthorized user")
         raise HTTPException(status_code=401, detail="Unauthorized")
     else:
-        return response.json()["id"], response.json()["role"]
+        return User(**response.json())
```
danscales, line 100: Note that the response from ResourceWatch has extra fields other than those in User, such as `_id`, `updatedAt`, and `provider`.  Will those extra keyword parameters to User work with the constructor of User, which is a pydantic model?  Just mentioning, in case that could be an issue.  But I see it seemed to be already working with the Signup class in the signup(() function in rw_api.py.

app/authentication/token.py
```diff
@@ -93,19 +88,34 @@ async def is_app_admin(token: str, app: str, error_str: str) -> bool:
...
-    """
+async def get_user(token: str = Depends(oauth2_scheme)) -> User:
+    """Get the details for authenticated user."""
 
     response: Response = await who_am_i(token)
 
-    if response.status_code == 401 or not (
-        "gfw" in response.json()["extraUserData"]["apps"]
-    ):
+    if response.status_code == 401:
```
danscales, line 96: Note that there are some existing callers of `get_user`  such as `create/get/delete_api_key` and you've relaxed the authentication on them, because you are no longer checking that gfw is in their list of apps.  Is that OK?

Might be good to note somewhere if that was intentional.

danscales (commented):
The new get_user and associated functions will make permission checking a lot easier, but see my comments, esp on the `applications` field.

app/authentication/token.py
jterry64: I think might be getting confused by that we have two types of "applications" now: https://resource-watch.github.io/doc-api/reference.html#application

I was assuming we'd be using the new one for the "data-api" application, but maybe I got that wrong. @solomon-negusse was that your understanding?

app/authentication/token.py
jterry64: I'm pretty sure it just ignore those extra parameters, and that's how it was already working, but good to note

app/authentication/token.py
jterry64: Good point. My understanding of our discussion the other day is that we think it's ok for any RW admin to also be an admin for Data API, and it shouldn't be tied to the "gfw" app anymore. I'll put in the "gfw" check back in though so we can merge this PR, we can discuss in huddle maybe and add a separate PR.

jterry64: Oh wait, I confused this with the admin check, didn't realizer we also don't let users even create an API key without the gfw app. That seems pretty unnecessary to me, especially since there's already non-GFW users. I think any user should be able to create and manage API keys

app/authentication/token.py
danscales: OK, but also note that the new 'applications' field is not available from the auth/check-logged call, if that is the one we want to use.

app/authentication/token.py
danscales: I'm fine with that.  I don't think you have to wait for the huddle on this - just maybe confirm directly with Solomon that is fine to remove the gfw app check here.

app/authentication/token.py
jterry64: ok on ready documentation, this `applications` is really just for RW API keys, we want to use `extraUserData.apps`. I think that's still confusing though because the possible applications in the docs all already use the Data API in some capacity: https://resource-watch.github.io/doc-api/concepts.html#existing-applications

So it feels we're changing the meaning of `application` to also list `data-api` as a separate application, that's used to manage resources for other applications. I think if we want to use the `application` field in the future, we should use it more like RW to control resource management on the Data API. Thoughts @danscales @dmannarino @solomon-negusse @gtempus?

PR #513, opened 2024-05-10
Check for owner on version updates

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Only admins are allowed to create/update versions.

Issue Number: GTC-2821

## What is the new behavior?

Allow dataset owners and admins to create/update versions.

- [ ] Yes
- [X] No

### DIFF
--- a/app/routes/datasets/dataset.py
+++ b/app/routes/datasets/dataset.py
@@ -28,12 +28,15 @@ router = APIRouter()
 async def get_owner(
     dataset: str = Depends(dataset_dependency), user: User = Depends(get_manager)
 ) -> User:
-    """Retrieves the user object that owns the dataset if that user is the one
-    making the request, otherwise raises a 401."""
+    """Returns the User making the request as long as that user is an admin or
+    the owner of the dataset, otherwise raises a 401."""
+
+    if user.role == "ADMIN":
+        return user
 
     dataset_row: ORMDataset = await datasets.get_dataset(dataset)
     owner: str = dataset_row.owner_id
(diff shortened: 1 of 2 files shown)


solomon-negusse (approved):
Looks neat

app/routes/datasets/dataset.py
```diff
@@ -33,7 +33,7 @@ async def get_owner(
 
     dataset_row: ORMDataset = await datasets.get_dataset(dataset)
     owner: str = dataset_row.owner_id
-    if owner != user.id:
+    if owner != user.id and user.role != "ADMIN":
```
dmannarino, line 36: Huh. I swear I had a check for the user role being admin above the get_dataset call, but something happened to it. Maybe that was the bool version from the other branch I ditched. Anyway, thanks for catching this. If we want to avoid an unnecessary DB query we can do that, though. Like
if user.role == "ADMIN":
    return user
Not a bid deal.

danscales: Yes, I noticed the same problem in my change.  As Daniel said, I put

if user.role == "ADMIN":
   return

above line 34

danscales (approved):
Looks good!

app/routes/datasets/dataset.py
jterry64: oh, I guess it didn't merge to the `data_manager` branch for some reason, I can add it before the DB

PR #514, opened 2024-05-10
Feature/data manager user fix

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

The pydantic model for the user doesn't match the response of `{rw-api}/auth/check-logged` response 

- [ ] Yes
- [x] No

### DIFF
--- a/app/models/pydantic/authentication.py
+++ b/app/models/pydantic/authentication.py
@@ -21,7 +21,7 @@ class User(StrictBaseModel):
     createdAt: datetime
     role: str
     provider: str
-    providerId: str
+    providerId: Optional[str]
     extraUserData: Dict[str, Any]
 
 
--- a/app/models/pydantic/authentication.py
+++ b/app/models/pydantic/authentication.py
@@ -20,7 +20,8 @@ class User(StrictBaseModel):
     email: EmailStr
     createdAt: datetime
     role: str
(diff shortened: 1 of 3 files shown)


dmannarino (approved):

app/models/pydantic/authentication.py
```diff
@@ -20,7 +20,8 @@ class User(StrictBaseModel):
     email: EmailStr
     createdAt: datetime
     role: str
-    applications: List[str]
+    provider: str
+    providerId: str
```
danscales, line 24: I don't see providerId listed in https://resource-watch.github.io/doc-api/reference.html#get-lt-base-api-url-gt-auth-check-logged .  Is that documentation out of date?  If it is not being returned by 'check-logged', can you put a comment in on why you're adding it?

Also, it looks like 'updatedAt' is returned by 'checked-logged'.  Do you want to include that for completeness?

solomon-negusse: I had one unpushed commit that makes providerId optional as it's only present when `provider` represents third party auth (facebook, google, etc). Thanks for catching it.
Also `updatedAt` is not present in the actual response.
The documentation looks incorrect/incomplete in both cases.

PR #515, opened 2024-05-10
GTC-2822 Check for dataset ownership on asset requests

danscales:
GTC-2822 Check for dataset ownership on asset requests

Make use of get_owner in the asset operations.

The checks for the pure asset operations were slightly trickier than dataset/version operations, because the dataset name is available in the URL path. You need to do a DB operation to get the dataset name. So, I put the actual get_owner() call in the main function rather than in a Depends() argument. I didn't try to add a new Depends() function that magically does the DB operation, since some of the functions already do the DB operation for other reasons.

Updated the test for update_metadata and update_field_metadata to test success and failure cases for different users.

Added a missing check for the ADMIN role in get_owner() and updated its comment.

Updated the detail of the various permission exceptions to be more informative (GTC-2795). Let me know what you think. We could put a bit more information in these messages if we did these errors in the main functions, rather than in the Depends functions, but that is probably not worth it.

In test_dataset.py, several tests were being skipped unintentionally because they were missing the '@pytyest.mark.asyncio' annotation, so I added that in where needed.

### DIFF
--- a/app/routes/assets/asset.py
+++ b/app/routes/assets/asset.py
@@ -71,6 +71,10 @@ from ...utils.path import infer_srid_from_grid
 from ..assets import asset_response
 from ..tasks import paginated_tasks_response, tasks_response
 
+from ...authentication.token import get_manager
+from ...models.pydantic.authentication import User
+from ..datasets.dataset import get_owner
+
 router = APIRouter()
 
 
@@ -103,12 +107,20 @@ async def update_asset(
     *,
     asset_id: UUID = Path(...),
     request: AssetUpdateIn,
-    is_authorized: bool = Depends(is_admin),
+    user: User = Depends(get_manager),
 ) -> AssetResponse:
     """Update Asset metadata.
 
     Only the dataset's owner or a user with `ADMIN` user role can do this operation.
     """
+
+    try:
+        asset_row: ORMAsset = await assets.get_asset(asset_id)
+    except RecordNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e))
+
+    # This is the actual check that the user is either the dataset owner or an admin
+    _ = await get_owner(asset_row.dataset, user)
 
     input_data = request.dict(exclude_none=True, by_alias=True)
 
@@ -133,7 +145,7 @@ async def update_asset(
 async def delete_asset(
     *,
     asset_id: UUID = Path(...),
-    is_authorized: bool = Depends(is_admin),
+    user: User = Depends(get_manager),
     background_tasks: BackgroundTasks,
 ) -> AssetResponse:
     """Delete selected asset.
@@ -148,6 +160,9 @@ async def delete_asset(
         row: ORMAsset = await assets.get_asset(asset_id)
     except RecordNotFoundError as e:
         raise HTTPException(status_code=404, detail=str(e))
+
+    # This is the actual check that the user is either the dataset owner or an admin
+    _ = await get_owner(row.dataset, user)
 
     if row.is_default:
         raise HTTPException(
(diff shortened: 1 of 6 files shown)


app/routes/datasets/dataset.py
```diff
@@ -28,13 +28,15 @@
 async def get_owner(
     dataset: str = Depends(dataset_dependency), user: User = Depends(get_manager)
 ) -> User:
-    """Retrieves the user object that owns the dataset if that user is the one
-    making the request, otherwise raises a 401."""
+    """Returns the User making the request as long as that user is an admin or
+    the owner of the dataset, otherwise raises a 401."""
 
+    if user.role == "ADMIN":
+        return user
```
dmannarino, line 35: Thanks for catching this, I thought I had it in the code but seem to have lost it. Justin caught it too.

tests_v2/unit/app/routes/datasets/test_dataset.py
```diff
@@ -102,16 +104,17 @@ async def test_get_owner_different_manager_fail(db, init_db, monkeypatch) -> Non
...
+@pytest.mark.asyncio
 async def test_get_owner_admin_success(db, init_db, monkeypatch) -> None:
     dataset_name: str = "my_first_dataset"
 
     from app.main import app
 
-    # Create a dataset
-    app.dependency_overrides[get_manager] = get_admin_mocked
+    # Create a dataset with a manager, then make sure get_owner succeeds with an admin.
+    app.dependency_overrides[get_manager] = get_manager_mocked
```
dmannarino, line 117: Thanks for adding the explanations!

dmannarino (approved):

app/routes/assets/asset.py
```diff
@@ -346,12 +361,22 @@ async def get_field_metadata(*, asset_id: UUID = Path(...), field_name: str):
...
     Only the dataset's owner or a user with `ADMIN` user role can do this operation.
     """
+
+    try:
+        asset_row: ORMAsset = await assets.get_asset(asset_id)
+    except RecordNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e))
+
+    # This is the actual check that the user is either the dataset owner or an admin
+    _ = await get_owner(asset_row.dataset, user)
```
jterry64, line 378: Is there a way we can get the asset row/dataset ID through a chained depends function, or do you think that's not preferable?

app/routes/datasets/dataset.py
jterry64: I'll make mine match this

app/routes/datasets/dataset.py
```diff
@@ -28,13 +28,15 @@
...
     dataset_row: ORMDataset = await datasets.get_dataset(dataset)
     owner: str = dataset_row.owner_id
     if owner != user.id:
-        raise HTTPException(status_code=401, detail="Unauthorized")
+        raise HTTPException(status_code=401, detail=f"Unauthorized write access to dataset {dataset} (or its versions/assets) by a user who is not an admin or owner of the dataset")
```
jterry64, line 39: Should we include actually the email of the owner? Otherwise they won't know who to contact. I added a function in another PR to look up a user by ID, I can add that part here

tests_v2/unit/app/routes/datasets/test_asset_metadata.py
```diff
@@ -42,12 +50,41 @@ async def test_update_asset_metadata(
...
-    resp = await async_client.patch(
-        f"asset/{tile_cache_asset.asset_id}/metadata", json={"max_zoom": max_zoom}
-    )
-
-    assert resp.json()["status"] == "success"
-    assert resp.json()["data"]["max_zoom"] == max_zoom
+    # generic_vector_source_version creates with owner of manager_mocked, so
+    # update by manager_mocked should succeed
+    appmain.dependency_overrides[get_user] = get_manager_mocked
+
```
jterry64, line 56: Not necessarily in this PR, but wondering if there's a way for us to turn this into a reusable helper function since we keep copying this code block

jterry64 (changes requested):

jterry64 (approved):

app/routes/assets/asset.py
danscales: As I discussed, I didn't want to do that, since some of the functions already fetch the asset row for their own reasons, so we would be doing the same DB access twice.

app/routes/datasets/dataset.py
danscales: I'll leave GTC-2795 open (better access denied error messages) and fix this in a later change, just so I can get the current change in today.

Thanks for the comment/suggestion!

PR #519, opened 2024-05-14
GTC-2788 Do proper serialization for UUIDs

danscales:
GTC-2788 Do proper serialization for UUIDs

This is in case there is a unusual query that does SELECT(*) on a dataset that has a UUID field. Currently, we get a 500 with an error "TypeError: Type is not JSON serializable: asyncpg.pgproto.pgproto.UUID".

We need to force UUID to str so it can be serialized, just as we are already doing with Decimal.

https://stackoverflow.com/questions/64395136/typeerror-object-of-type-uuid-is-not-json-serializable

I don't see any easy way to test via tests or in dev (need a dataset with UUID field in it), but will test on real datasets once this is in staging.


solomon-negusse (approved):

dmannarino (approved):
Nice, thanks for fixing this!

jterry64 (approved):

PR #525, opened 2024-05-21
GTC-2618 Allow appending layers to existing tables

manukala6:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: [GTC-2618](https://gfw.atlassian.net/browse/GTC-2618)

## What is the new behavior?

- Append layers or static files to existing tables on the API
- Remove deprecated status for append to version route
- Handle appends to test files (gfw-datapump updates) differently from other appends

- [ ] Yes
- [x] No

## Other information

[GTC-2618]: https://gfw.atlassian.net/browse/GTC-2618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
+            input_data["creation_options"]["layers"] = request.layers
+            
     background_tasks.add_task(
         append_default_asset, dataset, version, input_data, default_asset.asset_id
     )
 
     # We now want to append the new uris to the existing ones and update the asset
     update_data = {"creation_options": deepcopy(default_asset.creation_options)}
-    update_data["creation_options"]["source_uri"] += request.source_uri
+    update_data["creation_options"]["source_uri"] += request.source_uri # ERROR: only one source_uri is allowed
```
danscales, line 267: I don't understand the ERROR comment.  Are you saying that you should add a check that len(request.source_uri) == 1 ?  Should probably put that check in or remove the comment?

app/routes/datasets/versions.py
```diff
@@ -536,6 +559,17 @@ async def _version_response(
...
+#    with fiona.open(source_uri[0].replace("s3://", "/vsizip//vsis3/"), "r") as src:
+#        layers = src.layer_names
+#        for layer in layers:
+#            if layer in layers:
+#                return
+#            else:
+#                raise HTTPException(
+#                    status_code=400,
+#                    detail=f"Layer {layer} not found in source file."
+#                )
```
danscales, line 572: Should you get rid of this code if you're not going to use it?

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
 
     # We now want to append the new uris to the existing ones and update the asset
     update_data = {"creation_options": deepcopy(default_asset.creation_options)}
-    update_data["creation_options"]["source_uri"] += request.source_uri
+    update_data["creation_options"]["source_uri"] += request.source_uri # ERROR: only one source_uri is allowed
+    if input_data["creation_options"].get("layers") is not None:
+        if update_data["creation_options"]["layers"] is not None:
+            update_data["creation_options"]["layers"] += request.layers
+        else:
+            update_data["creation_options"]["layers"] = request.layers
```
danscales, line 272: It's confusing that this logic here is different from the logic above for input_data[].  Can you possibly add in a comment on what you expect the difference to be for input_data["creation_options"]["layers" and update_data["creation_options"]["layers"]?

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
         append_default_asset, dataset, version, input_data, default_asset.asset_id
     )
 
     # We now want to append the new uris to the existing ones and update the asset
     update_data = {"creation_options": deepcopy(default_asset.creation_options)}
-    update_data["creation_options"]["source_uri"] += request.source_uri
+    update_data["creation_options"]["source_uri"] += request.source_uri # ERROR: only one source_uri is allowed
+    if input_data["creation_options"].get("layers") is not None:
+        if update_data["creation_options"]["layers"] is not None:
+            update_data["creation_options"]["layers"] += request.layers
```
danscales, line 270: If request.layers is None (which you check for explicitly at line 256 above), then you will get a Python error here when you try add None (which is not a list) to an existing list.

terraform/vars/terraform-production.tfvars
```diff
@@ -5,6 +5,8 @@ rw_api_url                = "https://api.resourcewatch.org"
 desired_count             = 2
 auto_scaling_min_capacity = 2
 auto_scaling_max_capacity = 15
+fargate_cpu               = 2048
+fargate_memory            = 4096
```
danscales, line 9: I guess this is OK, but somehow you are including hotfixes from master into your change?

danscales (commented):
Are you missing some other test changes?  It seems like you added test.gpkg.zip, but then didn't use it.  Also, I was expecting another append test that explicitly tests your new code.

app/models/pydantic/creation_options.py
```diff
@@ -226,10 +226,8 @@ class VectorSourceCreationOptions(StrictBaseModel):
     def validate_source_uri(cls, v, values, **kwargs):
         if values.get("source_driver") == VectorDrivers.csv:
             assert len(v) >= 1, "CSV sources require at least one input file"
-        else:
-            assert (
-                len(v) == 1
-            ), "Non-CSV vector sources require one and only one input file"
+        elif values.get("source_driver") in [VectorDrivers.esrijson, VectorDrivers.shp, VectorDrivers.geojson_seq, VectorDrivers.geojson]:
+            assert (len(v) == 1), "GeoJSON and ESRI Shapefile vector sources require one and only one input file"
```
dmannarino, line 230: I think you're intending for this change to allow multiple input files for GPKG, but is that actually supported in the code? Maybe I'm looking in the wrong place, but in vector_source_assets.py it looks like we currently only make use of the first source_uri (perhaps because of the issues that might arise from specifying layers for multiple files?): https://github.com/wri/gfw-data-api/blob/master/app/tasks/vector_source_assets.py#L254

app/models/pydantic/versions.py
```diff
@@ -59,7 +60,14 @@ class VersionUpdateIn(StrictBaseModel):
...
     source_uri: List[str]
-
+    source_driver: Optional[VectorDrivers] = Field(
+        None, description="Driver of source file. Must be an OGR driver"
+    )
+    layers: Optional[List[str]] = Field(
+        None,
+        description="List of layer names to append to version. "
+        "If not set, all layers in source_uri will be appended.",
+    )
```
dmannarino, line 70: It looks to me like if layers is not specified it will be assumed that there is only a layer named like the file (though I may be looking in the wrong place in the code): https://github.com/wri/gfw-data-api/blob/master/app/tasks/vector_source_assets.py#L209-L214

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
     # For the background task, we only need the new source uri from the request
     input_data = {"creation_options": deepcopy(default_asset.creation_options)}
     input_data["creation_options"]["source_uri"] = request.source_uri
+
+    # If source_driver is "text", this is a datapump request
+    if input_data["creation_options"]["source_driver"] != "text":
```
solomon-negusse, line 245: Minor but would be good to use the enum here to make it future-proof `input_data["creation_options"]["source_driver"] != TableDrivers.text`

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
+        # Verify that source_driver is not None
+        if input_data["creation_options"]["source_driver"] is None:
```
solomon-negusse, line 247: The pydanic model for appending should be updated to take in `creation_options`, right? That'd be the right place to validate input like [so](https://github.com/wri/gfw-data-api/blob/master/app/models/pydantic/creation_options.py#L186):

solomon-negusse (changes requested):
@manukala6, i left some comments, mainly I'd encourage updating the append pydantic model to include creation options to handle validation.

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
+                detail="Source driver must be specified for non-datapump requests."
+            )
+
+        # Append the new layers to the existing ones
+        if input_data["creation_options"].get("layers") is None: # ERROR: layers is not defined
+            input_data["creation_options"]["layers"] = request.layers
+        elif request.layers is not None:
+            input_data["creation_options"]["layers"] += request.layers
+        else:
+            input_data["creation_options"]["layers"] = request.layers
```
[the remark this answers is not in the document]
manukala6: `input_data["creation_options"]` is the body for the append POST request, so we want to exclude the layers from the version creation creations for that operation. Ln 268-76 are updating the creation options object for the version itself. I added comments to reflect this.

app/routes/datasets/versions.py
manukala6: I've corrected this in later commits. Thanks for catching.

app/routes/datasets/versions.py
manukala6: I added comments to reflect this. `input_data["creation_options"]` is used for the append operation and `update_data["creation_options"]` is used to update the version's creation options to reflect changes made in the append request.

app/routes/datasets/versions.py
manukala6: I refactored this section to account for this. `None` is now not being added to a list.

app/models/pydantic/creation_options.py
manukala6: This wasn't for allowing multiple input files, rather its for updating the version creation options after the append operation is successful (there would be more than 1 source_uri in this case)

app/models/pydantic/versions.py
manukala6: Good catch, I corrected the description. Layer names are required for .GDB and .GPKG, otherwise the file name is used as the layer name.

app/routes/datasets/versions.py
```diff
@@ -240,13 +240,36 @@ async def append_to_version(
...
+    if input_data["creation_options"]["source_driver"] != "text":
+        # Verify that source_driver is not None
+        if input_data["creation_options"]["source_driver"] is None:
+            raise HTTPException(
+                status_code=400,
+                detail="Source driver must be specified for non-datapump requests."
+            )
+
+        # Append the new layers to the existing ones
+        if input_data["creation_options"].get("layers") is None: # ERROR: layers is not defined
```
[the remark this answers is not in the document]
manukala6: Thanks for the input, I refactored it similarly.

app/routes/datasets/versions.py
danscales: Great, thanks for the added comments to clarify!  The logic in both cases looks better now.

danscales (approved):

solomon-negusse (approved):

app/models/pydantic/creation_options.py
dmannarino: Okay, I see. But would it have the side effect of now allowing multiple GPKG sources to be specified, though those after the first will be silently ignored?

PR #528, opened 2024-05-24
Cache latest version per dataset

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- get_latest_version is called A LOT, and every time it's called it makes a DB call. 

Issue Number: N/A

## What is the new behavior?

- Cache the latest version value to reduce DB calls. Invalidate cache in new and update version functions, but set a TTL of 1 hour in case I missed something.

- [ ] Yes
- [x] No

### DIFF
--- a/app/crud/versions.py
+++ b/app/crud/versions.py
@@ -82,9 +82,6 @@ async def create_version(dataset: str, version: str, **data) -> ORMVersion:
     if data.get("is_downloadable") is None:
         data["is_downloadable"] = d.is_downloadable
 
-    if data.get("is_latest"):
-        await _reset_is_latest(dataset, version)
-
     metadata_data = data.pop("metadata", None)
     try:
         new_version: ORMVersion = await ORMVersion.create(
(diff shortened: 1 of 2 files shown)


app/crud/versions.py
```diff
@@ -100,6 +99,18 @@ async def create_version(dataset: str, version: str, **data) -> ORMVersion:
...
+    # latest on creation? I thought we didn't. Seems like it will cause
+    # requests to temporarily go to a perhaps incompletely-imported
+    # asset...
+    if data.get("is_latest"):
+        await _reset_is_latest(dataset, version)
+        # logger.info(
+        #     f"Setting version {version} to latest for dataset {dataset}. "
+        #     f"Cache info: {get_latest_version.cache_info()}"
+        # )
+        _: bool = get_latest_version.cache_invalidate(dataset)
```
danscales, line 112: Seems like it might make sense to  put the `cache_invalidate` call in `_reset_is_latest`, so we don't forget to call it when we call `reset_is_latest`?

app/crud/versions.py
```diff
@@ -155,6 +171,12 @@ async def _update_is_downloadable(
 
 
 async def _reset_is_latest(dataset: str, version: str) -> None:
+    """Set is_latest to False for all other versions of a dataset."""
+    # TODO: Should we make sure the version is valid to avoid setting nothing
+    # to latest? Or is being able to do that a desired feature?
```
danscales, line 176: I think you can remove this TODO, and just add a line to the header comment that says it is recommended to set is_latest flag for version before calling this method to reset is_latest for all other versions [since that is what is done in update_version].

app/crud/versions.py
```diff
@@ -100,6 +99,18 @@ async def create_version(dataset: str, version: str, **data) -> ORMVersion:
         )
         new_version.metadata = metadata
 
+    # FIXME: Should we really allow one to specify a new version as
+    # latest on creation? I thought we didn't. Seems like it will cause
+    # requests to temporarily go to a perhaps incompletely-imported
```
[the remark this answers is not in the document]
dmannarino: Ahh, okay, thanks for figuring that out! Will add the note.

app/crud/versions.py
dmannarino: Yeah, I'll do that. I was on the fence about it, but now that you say it, that seems better.

solomon-negusse (approved):

danscales (approved):

PR #532, opened 2024-05-28
allow users to specify non-globalforestwatch domains in api key

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Creating api key from non-globalforestwatch origins fails now if the user passes domains list that doesn't include globalforestwatch. 

## What is the new behavior?

- Users can create api key for non-GFW domains

- [ ] Yes
- [x] No

### DIFF
--- a/app/routes/authentication/authentication.py
+++ b/app/routes/authentication/authentication.py
@@ -74,16 +74,8 @@ async def create_api_key(
 
     input_data = api_key_data.dict(by_alias=True)
 
-    origin = request.headers.get("origin")
-    referrer = request.headers.get("referer")
-    if not api_key_is_valid(input_data["domains"], origin=origin, referrer=referrer):
-        raise HTTPException(
-            status_code=400,
-            detail="Domain name did not match the request origin or referrer.",
-        )
-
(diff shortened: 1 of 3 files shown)


dmannarino (approved):

app/routes/authentication/authentication.py
```diff
@@ -74,16 +74,8 @@ async def create_api_key(
 
     input_data = api_key_data.dict(by_alias=True)
 
-    origin = request.headers.get("origin")
-    referrer = request.headers.get("referer")
-    if not api_key_is_valid(input_data["domains"], origin=origin, referrer=referrer):
```
solomon-negusse, line 79: This prevents users from creating api keys from anywhere but the domain they plan to use the api key for (for example, one wouldn't be able to use postman to create an api key for mapbuilder (listed in the domains request body) because the origin or referer wouldn't point to that and this check will fail. 
This check get applied correctly checking requests coming in with an api key (https://github.com/wri/gfw-data-api/blob/master/app/authentication/api_keys.py#L53)

PR #539, opened 2024-06-06
Change github destroy action back to way it was in February.

danscales:
Change github destroy action back to way it was in February.

I think this should get the destroy operation running again (not skipped), when a feature branch is deleted.

### DIFF
--- a/.github/workflows/terraform_destroy_on_delete.yaml
+++ b/.github/workflows/terraform_destroy_on_delete.yaml
@@ -4,7 +4,7 @@ on: [delete]
 
 jobs:
   build:
-    if: contains(github.event.ref_type, 'branch') &&  (! github.event.ref == 'master') && (! github.event.ref == 'develop')
+    if: contains(github.event.ref_type, 'branch') &&  (! contains(github.event.ref, 'master')) && (! contains(github.event.ref, 'develop'))
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/terraform_destroy_on_delete.yaml
```diff
@@ -4,7 +4,7 @@ on: [delete]
 
 jobs:
   build:
-    if: contains(github.event.ref_type, 'branch') &&  (! github.event.ref == 'master') && (! github.event.ref == 'develop')
+    if: contains(github.event.ref_type, 'branch') &&  (! contains(github.event.ref, 'master')) && (! contains(github.event.ref, 'develop'))
```
dmannarino, line 7: Well, this would mean that branches with "master" or "develop" in their names would not be deleted (which was the original problem). It looks like the original has two problems: The leading exclamation mark is not the right way to specify inequality, and we're comparing to develop and master whereas I think in this case we should be comparing to the full path. How about changing to this line:
if: github.event.ref_type == 'branch' && (github.event.ref != 'refs/heads/master') && (github.event.ref != 'refs/heads/develop')

danscales: OK, thanks, made that change, and it makes sense to me now [like why we were using contains before, but it got it wrong for merge_from_master].   So, if you just want to approve this, I'll check this in and try testing it.

dmannarino (approved):
Great, thanks for taking care of this!

PR #545, opened 2024-06-12
Create Python script to export to GEE

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: GTC-2851

## What is the new behavior?

Add option to export COGs to GEE by created copy to GCS and creating a COG-backed asset using the GEE API. 

COGs are stored in the forma-250 GCS project in the bucket `data-api-gee-assets`. They have the naming convention `{dataset}/{implementation}.tif`. The GEE assets are in the forma-250 GEE project with the asset ID `{dataset}/{implementation}`.

I didn't implement versioning to start with, since that might quickly use up a lot of space in GCS/GEE and would be confusing to GEE users. So for now, assuming the asset is simply using the latest version in the API. I'm thinking we can revisit this in the future if there's a use case for better versioning.

- [ ] Yes
- [X] No

### DIFF
--- a/batch/python/export_to_gee.py
+++ b/batch/python/export_to_gee.py
@@ -39,8 +39,26 @@ def upload_cog_to_gcs(dataset, implementation):
 
 def create_cog_backed_asset(dataset, implementation, gcs_path, service_account):
     credentials = ee.ServiceAccountCredentials(service_account, GCS_CREDENTIALS_FILE)
-    session = AuthorizedSession(credentials.with_quota_project(EE_PROJECT))
+    ee.Initialize(credentials)
 
+    # delete any existing asset with the same dataset/implementatio
+    try:
+        ee.data.deleteAsset(f"projects/{EE_PROJECT}/assets/{dataset}/{implementation}")
+    except ee.EEException:
+        # asset doesn't exist
+        pass
+
+    # create dataset folder if it doesn't exist
+    try:
+        ee.data.createAsset(
+            {"type": "Folder"}, f"projects/{EE_PROJECT}/assets/{dataset}"
+        )
+    except ee.EEException:
+        # folder already exists
+        pass
+
+    # link GCS COG to the GEE asset
+    session = AuthorizedSession(credentials.with_quota_project(EE_PROJECT))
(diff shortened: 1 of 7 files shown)


batch/python/export_to_gee.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+
+
+def export_to_gee(
+    dataset: str = Option(..., help="Dataset name."),
+    implementation: str = Option(..., help="Implementation name."),
+):
+    service_account = set_google_application_credentials()
+    gcs_path = upload_cog_to_gcs(dataset, implementation)
+    create_cog_backed_asset(gcs_path, dataset, implementation, service_account)
```
danscales, line 65: This is calling create_cog_backed_asset() with arguments in the wrong order, I think?  [gcs_path should be 3rd argument, and you're not using keywords]

batch/scripts/get_arguments.sh
```diff
@@ -100,6 +100,11 @@ do
       shift # past argument
       shift # past value
       ;;
+      -G|--export_to_gee)
+      EXPORT_TO_GEE="$2"
+      shift # past argument
+      shift # past value
+      ;;
```
dmannarino, line 107: I think you just want 1 shift, and store something like "TRUE" (though I don't think the actual value matters). As is I think you'll be shifting past the next argument.

dmannarino: > I think you just want 1 shift, and store something like "TRUE" (though I don't think the actual value matters). As is I think you'll be shifting past the next argument.

If it's confusing, see the --overwrite example below.

batch/python/export_to_gee.py
jterry64: fixed

batch/scripts/get_arguments.sh
jterry64: oh good catch, fixed

danscales (approved):

dmannarino (approved):

solomon-negusse (approved):

PR #547, opened 2024-06-24
GTC-2868 Force use of only XY (2 dimensions) from the dataset

danscales:
GTC-2868 Force use of only XY (2 dimensions) from the dataset

Added "-dim xy" (same as "-dim 2") to the ogr2ogr command line.


danscales:
> Change looks good! How hard would it be to have a test for the CSV case at least?

Maanas indicated to me that the 3-dimension problem wasn't relevant for CSV or wasn't needed, so I didn't make any changes to `load_vector_csv_data.sh`.  So, can't do a test on that.  I tested this in dev using the dataset that Maanas gave to me (s3://gfw-data-lake-dev/landmark_icls_pt/v202406/raw/CommunityLevelData_pt.zip ) and will test again once this change is in staging.

manukala6 (approved):

danscales:
Thanks for the reviews!

PR #548, opened 2024-06-26
GTC-2864 Make test output for Data API be less verbose by default

danscales:
GTC-2864 Make test output for Data API be less verbose

I changed both script/test and script/test_v2 to not show warnings or
do/show code coverage by default. So, the actual test results won't
scroll off the screen in the output when running a single or few tests.
Also, I added an option to cause stdout of the tests to be displayed, so
that you can see debugging print() calls when running tests, even if the
test succeeds. (Normally all output of a test is captured and not
displayed if the test succeeds, which makes sense when not debugging.)
    
New options (all off by default):
  --do-cov - Run code coverage and display it after the test results
 --show-warning - Show python warnings when compiling tests
  --show-stdout - Show output of successful tests

Definitely feel free to comment on other ways to improve the Data API test output.

### DIFF
--- a/.github/workflows/terraform_build.yaml
+++ b/.github/workflows/terraform_build.yaml
@@ -12,8 +12,8 @@ jobs:
 
     - name: Test with pytest
       run: |
-        ./scripts/test
-        ./scripts/test_v2 --no_build
+        ./scripts/test --do-cov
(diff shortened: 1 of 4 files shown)


dmannarino (approved):
This is great, thanks!

PR #567, opened 2024-08-07
GTC-2920 Added documentation for the /query/batch route

danscales:
GTC-2920 Added some documentation for the /query/batch route

Including some documentation on the arguments.

Let me know of anything that you think I should add or describe in more detail.

### DIFF
--- a/app/models/pydantic/query.py
+++ b/app/models/pydantic/query.py
@@ -16,7 +16,7 @@ class QueryBatchRequestIn(StrictBaseModel):
         None, description="An inline collection of GeoJson features on which to do the same query"
     )
     uri: Optional[str] = Field(
-        None, description="URI to a CSV or GeoJson file giving a list of features on which to do the same query. For a CSV, the column with the geometry in WKB format should be named 'WKT' (not 'WKB')"
+        None, description="URI to a vector file in a variety of formats supported by Geopandas, including GeoJson and CSV format, giving a list of features on which to do the same query. For a CSV file, the column with the geometry in WKB format should be named 'WKT' (not 'WKB')"
     )
     id_field: str = Field(
         "fid", description="Name of field with the feature id, for use in labeling the results for each feature. This field must contain a unique value for each feature."
--- a/app/models/pydantic/query.py
+++ b/app/models/pydantic/query.py
@@ -3,6 +3,7 @@ from typing import Optional
 from app.models.enum.creation_options import Delimiters
 from app.models.pydantic.base import StrictBaseModel
 from app.models.pydantic.geostore import FeatureCollection, Geometry
+from pydantic import Field
 
 
 class QueryRequestIn(StrictBaseModel):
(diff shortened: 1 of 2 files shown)


app/routes/datasets/queries.py
```diff
@@ -336,8 +336,19 @@ async def query_dataset_list_post(
...
+    """Execute a READ-ONLY SQL query on the specified raster-based dataset version
+    for a potentially large list of features. The features may be specified by an
+    inline GeoJson feature collection or the URI of a file in CSV or GeoJson format.
+    For CSV files, the geometry column should be named "WKT" and the geometry values
+    should be in WKB format.
+
+    The specified sql query will be run on each individual feature, and so may take a
+    while. Therefore, the results of this query include a job_id. The user should
+    then periodically query the specified job via the /job/{job_id} api. When the
+    "data.status" indicates "saved", then the results will be available at the specified
```
jterry64, line 348: I think we changed the statuses a bit but you weren't there. Instead of "saved", its now "success" if everything succeeded, or "partial_success" if there were some failed geometries

app/routes/datasets/queries.py
```diff
@@ -336,8 +336,19 @@ async def query_dataset_list_post(
...
+    "data.download_link" and "data.failed_geometries_link".
+
+    """
```
jterry64, line 351: Should we include that we have a 5 minute timeout right now? That also should potentially be communicated in the status, or a "status_reason" field. We might need to add that later.

jterry64 (changes requested):

app/routes/datasets/queries.py
danscales: Good suggestion!  Added an extra sentence about the 5 minute timeout.

app/routes/datasets/queries.py
danscales: Good catch, I think I was looking at an older result before that change was made, but should have remembered it was changed.  Updated the text to describe "success", "partial_success", "failure", and "error".  As you mentioned, we probably have a bunch of future changes to improve the output messages to indicate the actually kind of "error", such as timeout or other step function failure.

app/models/pydantic/query.py
```diff
@@ -11,9 +12,15 @@ class QueryRequestIn(StrictBaseModel):
...
 class QueryBatchRequestIn(StrictBaseModel):
-    feature_collection: Optional[FeatureCollection]
-    uri: Optional[str]
-    id_field: str
+    feature_collection: Optional[FeatureCollection] = Field(
+        None, description="An inline collection of GeoJson features on which to do the same query"
+    )
+    uri: Optional[str] = Field(
+        None, description="URI to a CSV or GeoJson file giving a list of features on which to do the same query.  For a CSV, the column with the geometry in WKB format should be named 'WKT'"
+    )
```
dmannarino, line 20: I think technically many more file types are possible, no? It should be limited only by the preprocessing phase and the use of geopandas. I think it's written such that it will work with shapefiles, GPKGs, etc. Maybe you can write something vague (like "vector file") until we try all the possibilities?

app/models/pydantic/query.py
```diff
@@ -11,9 +12,15 @@ class QueryRequestIn(StrictBaseModel):
...
+    id_field: str = Field(
+        "fid", description="Name of field with the feature id, for use in labelling the results for each feature"
+    )
```
dmannarino, line 23: Maybe underscore the fact that the field should contain a unique value?

app/models/pydantic/query.py
dmannarino: Also, is this accurate: "the column with the geometry in WKB format should be named 'WKT'"
That seems... wrong. WKT != WKB.

danscales: OK, change the comments to refer to a vector file that is in any of a variety of formats supported by Geopandas, including GeoJson and CSV format.

Also, the column name does have to be WKT, even though the format is WKB..  I made that a bit clearer in the comments.

app/models/pydantic/query.py
danscales: Good suggestion, changed the text to make that clearer.  Thanks!

PR #586, opened 2024-09-23
on-demand ec2 instances for cog creation batch jobs

solomon-negusse:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

COG creation batch jobs run on SPOT instances that get terminated before completion

## What is the new behavior?

COG creation batch jobs run on on-demand instances

- [ ] Yes
- [x] No


app/models/pydantic/jobs.py
```diff
@@ -138,9 +139,9 @@ class PixETLJob(Job):
 
 
 class GDALCOGJob(Job):
-    """Use for creating COG files using GDAL Python docker in PixETL queue."""
+    """Use for creating COG files using GDAL Python docker in COG queue."""
 
-    job_queue = PIXETL_JOB_QUEUE
+    job_queue = COGIFY_JOB_QUEUE
```
jterry64, line 144: The terraform and code looks good, but I don't think we want to name this all specifically as "cogify". That's just one use case for this on demand queue, but we were also interested in using it eventually for other big jobs that often have issues on spot (e.g. like the TCL tile caches). 

So maybe just something like "on-demand-computing" or something? I guess we also have a "pixetl", we should label that more generically at some point too.

jterry64 (changes requested):

app/models/pydantic/jobs.py
solomon-negusse: Good call,  the pixetl one had thrown me off into thinking the queue is coupled with the specific compute environment.

danscales (approved):
Looks great!

dmannarino (approved):
Awesome!

PR #597, opened 2024-10-24
Sort the assets by creation time in the version endpoint, just like /assets

danscales:
Sort the assets by creation time in the version endpoint, just like the /assets endpoint

We made the asset list in the GET version endpoint more useful recently by including the asset-id. But that list is not being sorted by creation time, unlike the list provided by /assets, so the assets are somewhat randomly ordered.

So, add the change to sort the asset list of the GET version endpoint by creation time. This ensures that the default asset is always first, etc. (same as the ordering in _build_filtered_query)

### DIFF
--- a/app/routes/datasets/asset.py
+++ b/app/routes/datasets/asset.py
@@ -66,7 +66,8 @@ async def get_version_assets(
         description="The number of assets per page. Default is `10`.",
     ),
 ) -> Union[PaginatedAssetsResponse, AssetsResponse]:
-    """Get all assets for a given dataset version.
+    """Get all assets for a given dataset version. The list of assets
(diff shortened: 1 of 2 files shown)


dmannarino (approved):
Nice!

gtempus:
We don't know if this will break anyone, though, right? I'm not opposed to the change, but can we add an explanation of the `order_by` behavior in the docs, @danscales?

danscales:
> We don't know if this will break anyone, though, right? I'm not opposed to the change, but can we add an explanation of the `order_by` behavior in the docs, @danscales?

OK, added a comment for the docs for both the "Get Version" and "Get Version Assets" endpoints. Thanks for the suggestion!

PR #601, opened 2024-10-31
Add option to pre-warp to a common projection

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
```diff
@@ -67,13 +68,33 @@ async def raster_tile_set_asset(
...
+                dataset, creation_options.source_uri, target_prefix, target_crs, "unify_projection", callback
+            )
+        )
+
+        creation_options.source_uri = new_src_uris
+
+
+    jobs.append(
+        await create_pixetl_job(
+            dataset, version, creation_options, "create_raster_tile_set", callback
```
danscales, line 93: ***In the case the `unify_projection` is True, don't you need to specify a parent dependency argument here?  The `create_unify_projection_job` must run fully before the `create_pixetl_job` runs.

batch/scripts/unify_projection.sh
```diff
@@ -0,0 +1,50 @@
...
+  elif [[ $s == s3://* ]]; then
+    aws s3 cp --no-progress "$s" "$source_dir"
+  fi
+  echo "Done downloading $s to $source_dir"
+
+  reprojected_dir="REPROJECTED_${src_count}"
+  mkdir -p "$reprojected_dir"
+
+  cd $source_dir
+  for d in $((tree -dfi)); do
```
danscales, line 34: I think you mean $(tree -dfi) here (only one set of parens).  Also, I guess we are sure the docker has the 'tree' command in it?  You could use 'find' here as well.

I guess this may not matter right now, since our source URIs don't current have sub-directories of files.

batch/scripts/unify_projection.sh
```diff
@@ -0,0 +1,50 @@
...
+
+for s in ${SRC[@]}; do
+  source_dir="SRC_${src_count}"
+  mkdir -p "$source_dir"
+
+  echo "Now recursively downloading $s to $source_dir"
+  if [[ $s == gs://* ]]; then
+    gsutil cp -m -r "$s" "$source_dir"
+  elif [[ $s == s3://* ]]; then
+    aws s3 cp --no-progress "$s" "$source_dir"
```
danscales, line 26: I know we're not using aws for the source_uri, but shouldn't this have "--recursive" on it to copy an entire folder path and sub-folders?

danscales (commented):
Looks good so far (assuming xargs parallelization works well), but as I comment below, don't you need a dependency so the unify_projection job finishes completely before running the pixetl job?

dmannarino:
> Looks good so far (assuming xargs parallelization works well), but as I comment below, don't you need a dependency so the unify_projection job finishes completely before running the pixetl job?

You're right, thanks!

batch/scripts/unify_projection.sh
dmannarino: Good catch, thanks!

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
dmannarino: Yep, thanks!

batch/scripts/unify_projection.sh
dmannarino: I wasn't sure, I'll admit that was a try-it-and-see commit. It turns out we don't, and I've replaced the tree command with a find. Thanks for the parenthesis hint!

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
```diff
@@ -67,13 +68,42 @@ async def raster_tile_set_asset(
 
     creation_options = PixETLCreationOptions(**co)
 
+    jobs: List[Job] = list()
     callback: Callback = callback_constructor(asset_id)
 
-    create_raster_tile_set_job: Job = await create_pixetl_job(
-        dataset, version, creation_options, "create_raster_tile_set", callback
+    unify_job: Job | None = None
```
danscales, line 74: I think "Job | None" is more usually expressed as Optional[Job], probably more readable.

batch/scripts/_tiff_crosses_dateline.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+#
+# Small Script to check if input raster will
+# cross dateline when converting to EPSG:4326
+#
+# USAGE: ./crosses_dateline.sh infile [outfile]
+#
+# if no outfile is given, the script returns "true" or "false"
```
danscales, line 8: I think with the current modifications, there is no outfile arg, so maybe fix lines 6 and 8?

gtempus (commented):
I appreciate all the descriptions, @dmannarino . 👍 
No tests? 🤷 
`raster_tile_set_asset` is getting kinda long. Did you consider decomposing that function? Just curious.

app/tasks/raster_tile_set_assets/utils.py
```diff
@@ -225,3 +225,33 @@ async def create_resample_job(
...
+async def create_unify_projection_job(
+    dataset: str,
+    old_source_uris: List[str],
+    target_prefix: str,
+    target_crs: str,
+    job_name: str,
+    callback: Callback
+):
+    """
+    """
```
danscales, line 238: Can you add a function comment here?  Mainly would be that target_prefix is where the folders of warped files are placed, under SRC_0, SRC_1, SRC_2, etc.

danscales (approved):
Looks good!  Let's get it pushed into staging and try it out on the dist alerts composites!

app/tasks/raster_tile_set_assets/raster_tile_set_assets.py
dmannarino: If you think it's more readable, I'll change it! It's true that it had to be expressed as you say until Python 3.10. This is my first time actually using the new syntax.

batch/scripts/_tiff_crosses_dateline.sh
dmannarino: You're right, thanks!

dmannarino:
> I appreciate all the descriptions, @dmannarino . 👍 No tests? 🤷 `raster_tile_set_asset` is getting kinda long. Did you consider decomposing that function? Just curious.

True. Let me decompose and write a test as a follow-up.

batch/scripts/_tiff_crosses_dateline.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+ulx=$(jsonq "$gdalinfo" "['wgs84Extent']['coordinates'][0][0][0]")
+llx=$(jsonq "$gdalinfo" "['wgs84Extent']['coordinates'][0][1][0]")
+lrx=$(jsonq "$gdalinfo" "['wgs84Extent']['coordinates'][0][3][0]")
+urx=$(jsonq "$gdalinfo" "['wgs84Extent']['coordinates'][0][2][0]")
+
+crossing_dateline=false
+test $(python -c "print(${ulx}>${lrx})") = True && crossing_dateline=true
+test $(python -c "print(${ulx}>${lrx})") = True && crossing_dateline=true
```
solomon-negusse, line 38: @dmannarino: lines 37 and 38 look identical, just a duplicate or meant to check a different condition?

dmannarino: Oooh, good catch! It looks like I messed this up copying and pasting! They are meant to check a different condition. Will fix, thanks!

PR #608, opened 2024-11-22
Temporarily switch pixetl job queue to on-demand in staging only.

danscales:
Temporarily switch pixetl job queue to on-demand in staging only.

DO NOT move to production.


dmannarino (approved):

danscales:
Did this change and then later undid the change using a direct push.

PR #612, opened 2024-12-06
Introduce UV to replace Pipenv; multi-stage Dockerfile

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Using pipenv

Issue Number: N/A

## What is the new behavior?

Pipenv replaced with uv

- [ ] Yes
- [x] No


Dockerfile
```diff
@@ -1,43 +1,86 @@
...
+RUN --mount=type=cache,target=/app/.cache \
+    --mount=from=uv,source=/uv,target=./uv \
+    cd /_lock && \
+    /uv sync --locked --no-install-project
+
+# Start the runtime stage
+FROM ubuntu:noble
+SHELL ["sh", "-exc"]
+
+ENV PATH=$VIRTUAL_ENV/bin:/usr/local/bin:$PATH
```
danscales, line 55: Just checking - are all the environment variables (especially VIRTUAL_ENV) that were set during the build stage, also automatically set in the same way for the runtime stage?  I guess so, since otherwise tests wouldn't work, but it seems surprising, since you don't really want some of the build env variables set in the runtime docker.

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    "shapely",
+    "sqlalchemy<1.4",
+    "sqlalchemy-utils",
+    "starlette",
+    "typer",
+    "uvicorn[standard]",
+]
+
+[dependency-groups]
+dev = [
```
danscales, line 50: Seems like you dropped some packages under dev-packages in Pipfile, like asgi_lifespan, black, detect-secrets, and flake8.  Just checking, was that on purpose?

pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    "docker",
+    "geopandas",
```
danscales, line 52: You dropped some comments from the Pipfile for geopandas, gfw_pixetl, pandas, and retrying.  Maybe add those comments back her in this file?

danscales (commented):
Looks great! I assuming this is faster/more reliable/easier to maintain with uv than with pip/pipenv.

Just had a few questions/minor requests.

pyproject.toml
dmannarino: Good point, will restore them!

pyproject.toml
dmannarino: I've gone back and forth on this, but they are unnecessary. They are required by pre-commit, but it installs its own copy of them. This is true of everything but asgi-lifespan, which I seem to have moved to the main dependency section. It should stay in the dev dependencies, as it's only used for tests. I will move back asgi-lifespan, but leave out the pre-commit packages.

Dockerfile
dmannarino: Oh-HO! Thanks for asking! I've been confused about this question and finally looked it up. NO, the env variables do NOT persist across different stages! This is why I had to specify the pytest command the way I did, though I seem to remember being confused about it at the time.
That being said, I think I DO want some of them to persist, like the venv path. So I need to tweak this a little. Thanks again for asking!

dmannarino:
Closing in favor of new PR (branch name causes port collision in dev)

PR #616, opened 2024-12-23
GTC-3081: Add political/id-lookup endpoint

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

No geoencode endpoint

Issue Number: GTC-3081

## What is the new behavior?

- There is now an endpoint to look-up GADM v4.1 IDs by name

- [ ] Yes
- [x] No

## Other information

Open to suggestions for a better URL path. Not really crazy about either "thematic" or "geoencode".


danscales (commented):
Looks great!  Just a few minor comments

danscales:
> Looks great! Just a few minor comments

And great job with the tests!

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+
+@router.get(
+    "/geoencode",
+    tags=["Geoencoder"],
+    status_code=200,
+)
+async def geoencode(
+    *,
+    admin_source: str = Query(
+        "GADM", description="The source of administrative boundaries to use."
```
jterry64, line 24: Should we document that right now the only option is "GADM"?

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+    ),
+    admin_version: str = Query(
+        ...,
+        description="Version of the administrative boundaries dataset to use.",
```
jterry64, line 28: Similar, is there a way to document the choices available? I guess this may get more confusing if ever have multiple providers

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+            status_code=400,
+            detail=(
+                "Invalid admin boundary source. Valid sources:"
+                f" {[source for source in admin_source_to_dataset.keys()]}"
+            ),
+        )
+
+    version_str: str = "v" + str(admin_version).lstrip("v")
+
+    await version_is_valid(dataset, version_str)
```
jterry64, line 64: Like mentioned about the documentation above, should it be well known in advance which versions we support for providers, rather than just trying and throwing back an error?

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+    sql: str = _admin_boundary_lookup_sql(
+        adm_level, normalize_search, admin_source, *names
+    )
+
+    json_data: List[Dict[str, Any]] = await _query_dataset_json(
+        dataset, version_str, sql, None
+    )
+
+    return Response(
+        data={
```
jterry64, line 81: This might be cleaner as a pydantic model instead of a dict

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+            "adminSource": admin_source,
+            "adminVersion": admin_version,
+            "matches": [
+                {
+                    "country": {
+                        "id": match["gid_0"].rsplit("_")[0],
```
jterry64, line 87: A function for getting the ID here would be easier to read

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+                        "name": match["country"],
+                    },
+                    "region": {
+                        "id": (
+                            (match["gid_1"].rsplit("_")[0]).split(".")[1]
+                            if adm_level >= 1
```
jterry64, line 93: nitpick, I personally find ternary operators declared inside dicts confusing to read

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+                    "subregion": {
+                        "id": (
+                            (match["gid_2"].rsplit("_")[0]).split(".")[2]
+                            if adm_level >= 2
+                            else None
+                        ),
+                        "name": match["name_2"] if adm_level >= 2 else None,
+                    },
+                }
+                for match in json_data
```
jterry64, line 107: same with nested list comprehensions as above

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+async def version_is_valid(
+    dataset: str,
+    version: str,
+) -> None:
+    """Validate a version string for a given dataset."""
+    # Note: At some point I intend to change the version validator to
+    # use messaging like this. However, that is out of scope for the
+    # current ticket, and I want to see what people think, so I'll
+    # keep it here for now.
+    if re.match(VERSION_REGEX, version) is None:
```
jterry64, line 196: per above, maybe just have pre-validated versions rather than checking the version

jterry64 (changes requested):
Loving all the unit tests, had a few comments on style and documentation. Also, would possibly be able to update the PR with examples of requests/responses like Gary was doing for his GADM PRs? I found that quite helpful for assessing if the PR is actually doing what you're intending

app/routes/thematic/geoencoder.py
jterry64: Wild idea: should we consolidate admin version and admin dataset to one field, and have the options be like: "GADM 3.6", "GADM 4.1", "geoBoundaries 1.0", "middleEarth 3.2". Then it'll be clear what you're getting from a set of options.

gtempus: Combining the provider and version has merit (`middleEarth 3.2`). It has excellent cohesion.
This is such a simple scenario that it's hard to justify breaking it up.

However, we'd be encoding a string with special information (i.e., _provider_`space`_version_).
It's generally a good practice to split information up even though it makes for a bit more complexity.
It's more flexible in the long run (although I'm starting to violate YAGNI with this statement).
Finally, we are splitting this information up in the area/geostore microservice interface for all of the reasons mentioned above. 

That's my two cents.

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
+import re
```
jterry64, line 1: @dmannarino I know you said you didn't like thematic. I was thinking of other ideas, is this admin area specific endpoints, maybe we just put it under like something like "political"? E.g. `/political/geoencoder`. Then the future can include additional GADM endpoints, WDPA, concessions, etc. Not sure if we need to distinguish it from the rest of the API, or we can just have it all in under a header in the docs.

jterry64: But I do think `geoencoder` is a little vague if it only works for admin areas, geoencoding implies converting any text of the place into coordinates: https://en.wikipedia.org/wiki/Address_geocoding

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+def sanitize_names(
+    normalize_search: bool,
+    country: str | None,
+    region: str | None,
+    subregion: str | None,
+) -> List[str | None]:
+    """Turn any empty strings into Nones, enforces the admin level hierarchy,
+    and optionally unaccents and decapitalizes names.
+    """
+    names = []
```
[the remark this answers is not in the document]
dmannarino: Done, thanks!

app/routes/thematic/geoencoder.py
```diff
@@ -0,0 +1,216 @@
...
+        None,
+        description="Name of the region to match.",
+    ),
+    subregion: Optional[str] = Query(
+        None,
+        description="Name of the subregion to match.",
+    ),
+    normalize_search: bool = Query(
+        True,
+        description="Whether or not to perform a case- and accent-insensitive search.",
```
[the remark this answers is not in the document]
dmannarino: Think I need to even with the `default` argument set to True?

dmannarino:
> > Looks great! Just a few minor comments
> 
> And great job with the tests!

Thanks!

app/routes/thematic/geoencoder.py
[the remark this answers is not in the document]
danscales: OK, no, no change needed, I see now that the Default value (if not None) is displayed in the documentation.

app/routes/thematic/geoencoder.py
dmannarino: Done!

app/routes/thematic/geoencoder.py
dmannarino: I fundamentally agree that your suggestion is a good one, but it turns out to be difficult to do in practice. Especially considering the limited use of this endpoint and the fact that we will be omitting it from the docs.

app/routes/thematic/geoencoder.py
dmannarino: Changed to /political/id-lookup

app/routes/thematic/geoencoder.py
dmannarino: Done!

app/routes/thematic/geoencoder.py
dmannarino: I think they're all gone!

app/routes/thematic/geoencoder.py
dmannarino: Banished!

app/routes/thematic/geoencoder.py
dmannarino: Now specified on a per-env basis!

app/routes/political/id_lookup.py
```diff
@@ -0,0 +1,163 @@
...
+    json_data: List[Dict[str, Any]] = await _query_dataset_json(
+        dataset, version_str, sql, None
+    )
+
+    return form_geoencoder_response(
+        params.admin_source, params.admin_version, adm_level, json_data
+    )
+
+
+def sanitize_names(
```
gtempus, line 53: What about `normalilze_names`? Sanitize comes with a different connotation about what this method does.

gtempus (changes requested):
I like the new name of the endpoint, Daniel!

What about changing the name of the model from geoencoder? 😬

app/routes/political/id_lookup.py
dmannarino: Roger, will change, thanks!

dmannarino:
> I like the new name of the endpoint, Daniel!
> 
> What about changing the name of the model from geoencoder? 😬

Yes, that's the right thing to do. :) Will do, thanks!

app/routes/political/id_lookup.py
```diff
@@ -0,0 +1,166 @@
...
+        else:
+            subregion = {"id": extract_level_gid(2, match), "name": match["name_2"]}
+
+        matches.append({"country": country, "region": region, "subregion": subregion})
+
+    data = {
+        "adminSource": admin_source,
+        "adminVersion": admin_version,
+        "matches": matches,
+    }
```
danscales, line 159: Nit (optional): I guess this code works in Pydantic, but the types would be clearer if 'data' was a AdminIDLookupResponseData, which I guess you could do via:

data = AdminIDLookupResponseData(**{
  "adminSource": admin_source,
        "adminVersion": admin_version,
        "matches": matches,
  })

danscales (commented):
Looks good to me!

danscales (approved):

PR #617, opened 2025-01-02
GTC-3101 Add documentation for a bunch of operations and creation options

danscales:
GTC-3101 Add documentation for a bunch of operations and creation options

I put in doc lines for the most common options/operations that I use that are missing documentation.

Suggestions/corrections welcome!

If anyone can comment on the exact semantics of band_count and union_bands (and when they apply), I'd be happy to add documentation for those options as well.

### DIFF
--- a/app/models/pydantic/creation_options.py
+++ b/app/models/pydantic/creation_options.py
@@ -123,10 +123,11 @@ class RasterTileSetAssetCreationOptions(StrictBaseModel):
         )
     )
     pixel_meaning: str = Field(
-        ..., description="Short description of what the pixel value in the "
+        ..., description="Description of what the pixel value in the "
         "raster represents. This is used to clarify the meaning of the raster "
         "and distinguish multiple raster tile sets based on the same dataset "
-        "version."
+        "version. The pixel_meaning string should be fairly short, use all "
+        "lower-case letters, and use underscores instead of spaces."
     )
     data_type: DataType
     nbits: Optional[int] = Field(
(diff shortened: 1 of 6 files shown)


app/models/pydantic/creation_options.py
```diff
@@ -122,24 +122,60 @@ class RasterTileSetAssetCreationOptions(StrictBaseModel):
             "when input files are in different projections from each other."
         )
     )
-    pixel_meaning: str
+    pixel_meaning: str = Field(
```
jterry64, line 125: Should you specify it has to lowercase, spaces instead of underscores, not too long, etc.? Right now you'll just get some regex error response, might be useful to state it here.

app/models/pydantic/creation_options.py
```diff
@@ -122,24 +122,60 @@ class RasterTileSetAssetCreationOptions(StrictBaseModel):
...
+        ..., description="Short description of what the pixel value in the "
+        "raster represents. This is used to clarify the meaning of the raster "
+        "and distinguish multiple raster tile sets based on the same dataset "
+        "version."
+    )
     data_type: DataType
-    nbits: Optional[int]
-    calc: Optional[str]
+    nbits: Optional[int] = Field(
```
jterry64, line 132: Is it possible to link to the documentation from GDAL in these?

jterry64 (approved):
Some small comments but otherwise looks good, thanks for doing this!

app/models/pydantic/creation_options.py
```diff
@@ -122,24 +122,60 @@ class RasterTileSetAssetCreationOptions(StrictBaseModel):
...
+        None,
+        description="For raster sources or default assets, a raster algebra "
+        "expression, similar to gdal_calc, to "
+        "combine multiple raster layers into a new layer. Each source is "
+        "treated as a separate band, and is assigned a letter variable A-Z "
+        "based on the order in the sources. This expression can technically be "
+        "any valid NumPy expression.\n\nFor vector default assets, an SQL "
+        "expression specifying a calculation that yields the desired raster "
+        "value based on the fields of your vector dataset."
+    )
```
dmannarino, line 147: What do you think about this reorganization of the text? I tried to make the two different modes more explicit and clarify the part about band variable assignment:
  "There are two modes for this field, one for rasterizing vector sources and one for transforming one or more sources that are already raster. For rasterizing vector sources, this field should be an SQL expression that yields the desired raster value (presumably) based on the fields of your vector dataset. For raster sources, this should be a raster algebra expression, similar to that provided to gdal_calc, that transforms one or more input bands into one or more output bands. For use in this expression, each band in the sources is assigned an alphabetic variable (A-Z, then AA-AZ, etc.) in the order it exists in those sources, with those of the first source first, continuing with those of the second, and so on. So with two input sources of two bands each, they would be assigned to variables A and B (for the first source) and C and D (for the second source). NumPy is in scope, as np.

dmannarino:
This is wonderful, thanks for doing it, Dan!

dmannarino (approved):

app/models/pydantic/creation_options.py
danscales: Done!

app/models/pydantic/creation_options.py
danscales: Done!

app/models/pydantic/creation_options.py
danscales: Done, took your suggested reorganization, with just slight modifications.  Thanks!

PR #621, opened 2025-01-17
GTC-3084: Add new endpoints that proxy to geostore microservice (for now)

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: GTC-3084

## What is the new behavior?

- There are new routes which mimic and proxy to those of the RW geostore microservice, with Pydantic models to support them. The only mildly complicated routes are those to that conflict with previously existing GFW Data API routes, specifically creating a geostore and those getting a geostore by ID. Fortunately for the former the payload differs, and in the case of the latter the IDs are slightly different (GUIDs for RW, UUIDs for GFW). Tests have been added to ensure that in these cases the correct branch/function is chosen, though there are not comprehensive tests for the purely proxied routes.

- [ ] Yes
- [x] No

## Other information

Gary made a great ticket with more context which can be found here: https://gfw.atlassian.net/browse/GTC-3084


app/utils/rw_api.py
```diff
@@ -185,3 +195,210 @@ async def signup(name: str, email: str) -> User:
         )
 
     return User(**response.json()["data"])
+
+
+async def create_rw_geostore(
+    payload: RWGeostoreIn, x_api_key: str | None = None
+) -> RWGeostoreResponse:
+    url = f"{RW_API_URL}/v1/geostore"
```
gtempus, line 203: This should be `v2`, right?
```suggestion
    url = f"{RW_API_URL}/v2/geostore"
```

app/utils/rw_api.py
```diff
@@ -185,3 +195,210 @@ async def signup(name: str, email: str) -> User:
...
+    if response.status_code == 200:
+        return RWGeostoreResponse.parse_obj(response.json())
+    else:
+        raise HTTPException(response.status_code, response.text)
+
+
+async def calc_area(
+    payload: Dict, x_api_key: str | None = None
+) -> RWCalcAreaForGeostoreResponse:
+    url = f"{RW_API_URL}/v1/geostore/area"
```
gtempus, line 241: This should be `v2`, right?
```suggestion
    url = f"{RW_API_URL}/v2/geostore/area"
```

app/routes/geostore/geostore.py
```diff
@@ -1,52 +1,267 @@
...
-    try:
-        result: Geostore = await geostore.get_gfw_geostore_from_any_dataset(geostore_id)
-    except RecordNotFoundError as e:
-        raise HTTPException(status_code=404, detail=str(e))
+    dataset. If the provided ID is in UUID style, get from the GFW Data API.
+    Otherwise, forward request to RW API.
+    """
+    # Checking for hyphens to see if the provided ID is a GUID or a UUID
+    # is a bit iffy. Happy to hear better ideas.
+    if "-" in geostore_id:
```
gtempus, line 105: Good question. RW API MongoDB uses `ObjectId` (24 hex digits). Luckily, that's different from the postgres `uuid` (32 hex digits plus hyphens).

One example could be:
```python
from bson.objectid import ObjectId
import uuid

def check_id_type(value):
    try:
        ObjectId(value)
        return "ObjectId"
    except (TypeError, ValueError):
        try:
            uuid.UUID(value)
            return "UUID"
        except ValueError:
            return "Neither ObjectId nor UUID"

# Example usage
value1 = "60c0a4754483e60015000000"
value2 = "c78f40f4-3516-4099-8676-85507701973f"
value3 = "not an id"

print(check_id_type(value1))  # Output: ObjectId
print(check_id_type(value2))  # Output: UUID
print(check_id_type(value3))  # Output: Neither ObjectId nor UUID
```
`bson` is in the _pymongo_ [package](https://pymongo.readthedocs.io/en/stable/api/bson/objectid.html)

app/utils/rw_api.py
```diff
@@ -185,3 +195,210 @@ async def signup(name: str, email: str) -> User:
...
+) -> RWGeostoreResponse:
+    url = f"{RW_API_URL}/v2/geostore/{geostore_id}"
+
+    async with AsyncClient() as client:
+        if x_api_key is not None:
+            response: HTTPXResponse = await client.get(
+                url, headers={"x-api-key": x_api_key}
+            )
+        else:
+            response = await client.get(url)
```
gtempus, line 230: Did you experiment with generalizing this at all to cut down on duplication? Just curious?

gtempus (changes requested):
Nice work on getting tests worked in here, too! 💪 
I just had some basic questions/ideas for ya.

Thanks, @dmannarino!

app/utils/rw_api.py
dmannarino: > This should be `v2`, right?

Hmm, I've never tried with v2. The original ticket didn't say, and I got the v1 from https://resource-watch.github.io/doc-api/reference.html#overview-of-available-endpoints

I just tried, and they both work. Seemingly identical results.

app/utils/rw_api.py
dmannarino: > This should be `v2`, right?

Same as the create geostore endpoint (v1 and v2 both work).

app/utils/rw_api.py
gtempus: Let's stick with your current implementation, @dmannarino. We want bug-for-bug compatibility right now.

I looked at the code for the above `v1` endpoints and compared with `v2`. They are different.
But, I looked at Flagship's usage of the `create` endpoint. It looks like it just calls `/geostore` with no version and that seems to make the microservice default to `v1`. I could not find evidence of `calculate area` being called.

We'll need to reconcile the `v1` vs `v2` endpoints at a later date.

dmannarino: Okay, thanks for checking that out!

app/routes/geostore/geostore.py
dmannarino: Thanks for the info, and code snippet!

app/utils/rw_api.py
gtempus: Resolved in previous comment: https://github.com/wri/gfw-data-api/pull/621#discussion_r1925501964

app/routes/geostore/geostore.py
dmannarino: Hmm, it sure looks like the get geostore endpoint is not using ObjectIds. I have tried instantiating an ObjectId with a bunch of valid identifiers and always get
`ObjectId("ca38fa80a4ffa9ac6217a7e0bf71e6df")
InvalidId: 'ca38fa80a4ffa9ac6217a7e0bf71e6df' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string
`

If I try turning it into a bytes object first with unhexlify I get this rather confusing error:
`In [17]: foo = unhexlify("ca38fa80a4ffa9ac6217a7e0bf71e6df")
In [18]: ObjectId(foo)

TypeError: id must be an instance of (bytes, str, ObjectId), not <class 'bytes'>
`

app/utils/rw_api.py
dmannarino: I thought about writing a proxy_to_rw function to reduce duplication, but I wanted to see what they looked like first. And then I ran out of time. However I also suspected that a proxy function would lose value as we started modifying the routes to extend them. I tried to keep the pattern similar though, to make refactoring in the future easy.

jterry64:
@dmannarino @gtempus this PR is a little surprising to me, maybe I didn't fully understand the needs for this proxy. What are all the places this will need to be used? Will these endpoints be public? It does feel like we're suddenly adding the API structure of RW to Data API. What do we do to support endpoints like land use (which are likely very out of date) if they need to be updated? Maybe discuss in standup.

app/routes/geostore/geostore.py
gtempus: My bad! 🤦 I was thinking about the actual document ID this whole time and not the `hash`. It doesn't help that the `gfw-geostore-api` source refers to the hash as an ID frequently.

You're totally right, both are `uuid`s.

app/utils/rw_api.py
danscales: It seems like you should probably make a common function for all the get proxies out of lines 220-231, excluding line 229 (as Gary is asking)  That is, the common function issues the request with api-key or not, and throws an error if status code is not 200.  The only difference between all the GET proxies is the URL and the response type.

danscales (approved):

app/models/enum/geostore.py
```diff
@@ -4,3 +4,18 @@
...
+    # mining = "mining"  # Present in the docs, but yields a 404
+    oilpalm = "oilpalm"
+    tiger_conservation_landscapes = "tiger_conservation_landscapes"
+
+
+class LandUseTypeUseString(str, Enum):
+    fiber = "gfw_wood_fiber"
+    logging = "gfw_logging"
+    oilpalm = "gfw_oil_palm"
+    tiger_conservation_landscapes = "tcl"
```
gtempus: Having `enum`s for land use will not work since the param is [open-ended in the gfw-geostore-api microservice](https://github.com/gfw-api/gfw-geostore-api/blob/a6cc3432c90d316106af8af7e90c362f9cc5d7b8/app/src/routes/api/v2/geoStore.router.js#L219).

I quick look at the geostore logs for the past 10 days shows that `default` is frequently used with values such as:
- `river_basins`
- `biodiversity_hotspots`
- `endemic_bird_areas`

to name a few.

dmannarino: Thanks for re-discovering/remembering this! It's fixed now, I'm just passing through the string for land_use as-is. Just tested with these examples in dev!

PR #622, opened 2025-01-22
GTC-3125 Data API changes to support geostore ids for batch queries

danscales:
GTC-3125 Data API changes to support geostore ids for batch queries

We are just adding the 'geostore_ids' field to the list query request, and passing it through to the list query step function. Updated the check to make sure only one of 'feature_collection', 'uri', and 'geostore_ids' is specified.

My corresponding changes for the raster-analysis module (GTC-2965) are in a separate PR that is already checked into develop/staging.

### DIFF
--- a/app/models/pydantic/query.py
+++ b/app/models/pydantic/query.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, List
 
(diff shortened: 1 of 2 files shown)


solomon-negusse (approved):
Looks good to me!

PR #624, opened 2025-01-29
Introduce UV to replace Pipenv; multi-stage Dockerfile (new PR)

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Using pipenv

Issue Number: N/A

## What is the new behavior?

Pipenv replaced with uv

- [ ] Yes
- [x] No


Dockerfile
```diff
@@ -1,43 +1,88 @@
...
 
-# Install python dependencies
-# Install everything for dev and test otherwise just core dependencies
-COPY Pipfile Pipfile
-COPY Pipfile.lock Pipfile.lock
+# Create a virtual environment with uv inside the container
+RUN curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | sh && \
+    uv venv ${VENV_DIR} --python ${PYTHON_VERSION} --seed
 
-RUN if [ "$ENV" = "dev" ] || [ "$ENV" = "test" ]; then \
```
danscales, line 23: I don't see any equivalent checks on the right to install all dependencies or only production dependencies.  Is that functionality here somewhere? If not, I think you need to add it before checking in, right?  I do see the 'dev' dependency group in the pyproject.toml.

danscales (commented):
Can you add directions in a README as to how someone would install and use uv to install the data-api .venv in their local tree (for debugging, etc.)?  We wouldn't be able to use pipenv anymore, since the Pipfile is gone (as expected, since you don't want to maintain both).

Dockerfile
dmannarino: Installing dev deps in staging and prod fixed, thanks!

dmannarino:
> Can you add directions in a README as to how someone would install and use uv to install the data-api .venv in their local tree (for debugging, etc.)? We wouldn't be able to use pipenv anymore, since the Pipfile is gone (as expected, since you don't want to maintain both).

Will do!

solomon-negusse (approved):
Thank you! 🎉

danscales (approved):

scripts/setup
```diff
@@ -2,8 +2,34 @@
 
 set -e
 
-pip3.8 install pipenv
-/Library/Frameworks/Python.framework/Versions/3.8/bin/pipenv install --dev
+PYTHON_VERSION="3.10"
+VIRTUAL_ENV="${VIRTUAL_ENV:-.venv_uv}"
 
-detect-secrets scan > .secrets.baseline
+# Install Python and create the venv all in one go
```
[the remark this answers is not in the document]
dmannarino: I'm reluctant to install via an arbitrary method, but I'll add a check for uv in the path, and a link to the installation instructions. Thanks!

PR #632, opened 2025-02-12
Cleanup models

loganbyers:
Suggesting some cleanup around the models and enums...

### Replace sqlalchemy with vanilla enum
Pretty confident this will be without issue based on the way I have seen it used in the backend codebase... Really used as a way to get a limited set of strings... But worth pointing out that while the sqlalchemy Enum has typing similarity to enum.Enum they also have differences...

### Use relative imports everywhere
Across the entire `app/models/` directory there are inconsistent ways of performing imports sometimes looking at relative paths and othertimes looking from the root of the app. Under the assumption that there is no benefit/feature being utilized when performing an "absolute import" all imports have been reshaped to be relative imports.

### Remove PGType from creation_options enum file
There were two instances of `PGType` - both of type `Enum` but defined in different files. The one that is removed seems redundant or dead... There was a slight difference in the set of enum options but based on my searching of the codebase it is fine to remove the one under the `creation_options.py` file.

### DIFF
--- a/app/models/enum/creation_options.py
+++ b/app/models/enum/creation_options.py
@@ -61,24 +61,6 @@ class PartitionType(str, Enum):
     range = "range"
 
 
-class PGType(str, Enum):
-    bigint = "bigint"
-    boolean = "boolean"
-    character_varying = "character varying"
-    date = "date"
-    double_precision = "double precision"
-    geometry = "geometry"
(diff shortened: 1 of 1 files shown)


dmannarino (approved):
This looks great, thanks for doing it!

PR #633, opened 2025-02-18
Create mock endpoints for new drivers datamart endpoints

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Drivers data is pulled from a mix of precalculated tables and OTF.

Issue Number: GTC-3114

## What is the new behavior?
Create endpoints using the new DataMart model. See GTC-3114 for advanced details on the requests/responses and request flow. This PR only implements mocks that follow the structure of the request flow from an outside perspective.

- [ ] Yes
- [X] No


app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+@router.get(
+    "/tree-cover-loss-by-driver",
+    response_class=ORJSONResponse,
+    response_model=Response,
+    tags=["Land"],
+)
+async def tree_cover_loss_by_driver_search(
+    *,
+    geostore_id: UUID = Query(..., title="Geostore ID"),
+    canopy_cover: int = Query(..., alias="canopy_cover", title="Canopy Cover Percent"),
```
dmannarino, line 29: Should we have a default here? Don't we default elsewhere to 30 or 35 percent I think?

app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+    if not hasattr(request.app.state, "mock_ids"):
+        request.app.state.mock_ids = {}
+
+    try:
+        resource_id = request.app.state.mock_ids[f"{geostore_id}_{canopy_cover}"]["id"]
+        return ORJSONResponse(
+            status_code=200,
+            content={
+                "status": "success",
+                "data": {"link": f"/v0/land/tree-cover-loss-by-driver/{resource_id}"},
```
dmannarino, line 44: Should we make this a proper fully-qualified link? With the host etc.?

app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+
+@router.get(
+    "/tree-cover-loss-by-driver/{uuid}",
+    response_class=ORJSONResponse,
+    response_model=Response,
+    tags=["Land"],
+)
+async def tree_cover_loss_by_driver_get(
+    *,
+    uuid: UUID = Path(..., title="Tree cover loss by driver ID"),
```
dmannarino, line 66: Could we name this something more descriptive? Resource ID? Job ID? Also the title is a little confusing. At first I thought this was a UUID assigned to a driver category.

app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+    resource = None
+    for mock_id in request.app.state.mock_ids.values():
+        if mock_id["id"] == uuid:
+            resource = mock_id
+
+    if resource is None:
+        return ORJSONResponse(
+            status_code=404,
+            content={
+                "status": "status",
```
dmannarino, line 86: I think you meant to set this to failed.

app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+                "message": "Not Found",
+            },
+        )
+
+    if resource["retries"] < 3:
```
dmannarino, line 91: Hmm. Isn't resource a UUID? Or at least a str containing a UUID? It looks like you are thinking resource is a Dictionary.

app/routes/datamart/land.py
```diff
@@ -0,0 +1,171 @@
...
+        resource["retries"] += 1
+        return ORJSONResponse(
+            status_code=200,
+            headers={"Retry-After": "1"},
+            content={"data": {"status": "pending"}, "status": "success"},
+        )
```
dmannarino, line 97: Will this ever converge/be marked finished? I would think instead you would do something like 
`if resource["retries"] >= 1:
     resource["retries"] -= 1`

app/routes/datamart/land.py
dmannarino: D'oh, sorry, mis-read this. Ignore.

solomon-negusse (approved):
LGTM but let's remember to add the endpoints to api gateway terraform code in https://github.com/wri/gfw-data-api/blob/master/terraform/modules/api_gateway/gateway/main.tf

app/routes/datamart/land.py
jterry64: @dmannarino I was unsure about too, I've seen APIs do both. It looks like RW gives fully qualified links, so I'll go with that for consistency

app/routes/datamart/land.py
jterry64: @dmannarino I think I have right, it converges always when I've tested. The retries is set to a random number from 0-3 when created, and so it needs to be incremented. Once it's past that, it'll return below to give the full resource.

dmannarino (approved):

app/routes/datamart/land.py
gtempus: Yes, I think we should have fully qualified links, too.

gtempus (commented):
I like it! Seems to be shaping up really nicely, @jterry64. 💯

PR #635, opened 2025-02-25
Drivers OTF backend

jterry64:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: GTC-3167

## What is the new behavior?

- Now call OTF query service to compute results in a background task from POST
- POST immediately stores initial "pending" status in file and GET checks that for whether to use Retry-After
- Compute background task stores results in a file
- GET requests read results from file
- Change tests to use mocks instead of full end-to-end

- [X] Yes
- [ ] No

Slight changes to the model (mostly the metadata) of things I realized while working on it. This will be updated in the ticket GTC-3114.


app/routes/datamart/land.py
```diff
@@ -131,20 +84,33 @@ async def tree_cover_loss_by_driver_get(
 )
 async def tree_cover_loss_by_driver_post(
     data: TreeCoverLossByDriverIn,
+    background_tasks: BackgroundTasks,
```
danscales, line 87: Is the deprecated=True at line 83 intended, or maybe just a copy/paste error from somewhere else?

app/routes/datamart/land.py
```diff
@@ -131,20 +84,33 @@ async def tree_cover_loss_by_driver_get(
...
     api_key: APIKey = Depends(get_api_key),
 ):
     """Create new tree cover loss by drivers resource for a given geostore and
     canopy cover."""
 
+    # check geostore is valid
+    try:
+        await get_geostore(data.geostore_id, GeostoreOrigin.rw)
```
danscales, line 95: Not sure if this is worth fixing, but this get_geostore() call access checks both rw and gfw geostores, so it will succeed for a gfw geostore id.  But in compute_tree_cover_loss_by_driver, the get_geostore() calls requires assumes the geostore is from RW, will fail if it is from GFW, I think.

app/routes/datamart/land.py
jterry64: oh, copy paste error

app/routes/datamart/land.py
jterry64: Oh, I may have misunderstood this, I thought this would also check gfw too. I want to check both, will double check how we normally do that.

danscales: Actually, ignore this comment, I got confused.  You're using the same get_geostore() call in both, so you will check both geostores in both cases, and there's no inconsistency.

solomon-negusse (approved):

app/routes/datamart/land.py
```diff
@@ -66,85 +73,62 @@ async def tree_cover_loss_by_driver_get(
...
+        raise HTTPException(
+            status_code=422,
+            detail=f"Geostore {data.geostore_id} can't be found or is not valid.",
+        )
+
     # create initial Job item as pending
     # trigger background task to create item
     # return 202 accepted
-    resource_id = _get_resource_id(data.geostore_id, data.canopy_cover)
+    dataset_version = DEFAULT_LAND_DATASET_VERSIONS | data.dataset_version
```
dmannarino, line 115: At first I thought you were using a default OR the one provided, and thought you had it the wrong way round. But this is actually merging two dictionaries, right?

app/routes/datamart/land.py
```diff
@@ -66,85 +73,62 @@ async def tree_cover_loss_by_driver_get(
...
+
+    headers = {}
+    if resource["status"] == "pending":
+        headers = {"Retry-After": "1"}
+
+    return ORJSONResponse(
+        status_code=200,
+        headers=headers,
+        content={"data": resource, "status": "success"},
+    )
```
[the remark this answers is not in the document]
jterry64: I think response model would make sense. In fact, I think we already have some JSEND wrapper I should be using?

app/routes/datamart/land.py
jterry64: Yes! New syntax in Python 3.9+ I just learned about, I think I got the ordering right based on my test results... basically merges two dicts, with the one on the right overwriting any clashing keys

app/routes/datamart/land.py
jterry64: @danscales removed

jterry64:
@danscales @dmannarino I think I resolved all your comments if you want to take a quick look again. Main change is I switched to using consistent pydantic models everywhere, and also added the ability to specify a dataset/version.

app/models/orm/queries/raster_assets.py
```diff
@@ -1,7 +1,4 @@
-from ....application import db
-from ...enum.pixetl import Grid
-
-sql = """
+latest_raster_tile_sets = """
```
danscales, line 1: Can you maybe change the name of this variable, since it is now all versions of all raster tile sets (on the right grid), rather than just the latest versions?

app/models/pydantic/datamart.py
```diff
@@ -1,8 +1,49 @@
...
+
+class DataMartMetadata(StrictBaseModel):
+    geostore_id: UUID
+    sources: list[DataMartSource]
+
+
+class DataMartResource(StrictBaseModel):
+    status: str = "saved"
+    details: Optional[str] = None
+    metadata: DataMartMetadata = None
```
danscales, line 22: Change DataMartMetadata -> Optional[DataMartMetadata] to make typing more correct.

app/routes/datasets/queries.py
```diff
@@ -449,7 +459,9 @@ async def _query_dataset_json(
         return await _query_table(dataset, version, sql, geometry)
     elif query_type == QueryType.raster:
         geostore = cast(GeostoreCommon, geostore)
-        results = await _query_raster(dataset, default_asset, sql, geostore)
+        results = await _query_raster(
+            dataset, default_asset, sql, geostore, raster_version_overrides
```
danscales, line 463: **You need to specify "version_overrides=raster_version_overrides", since there are other omitted parameters format and delimiter before version_overrides.

app/routes/datasets/queries.py
```diff
@@ -811,10 +827,14 @@ def _get_default_layer(dataset, pixel_meaning):
...
+@hash_dict
 @alru_cache(maxsize=16, ttl=300.0)
-async def _get_data_environment(grid: Grid) -> DataEnvironment:
+async def _get_data_environment(
+    grid: Grid, version_overrides: Dict[str, str] = {}
+) -> DataEnvironment:
     # get all raster tile set assets with the same grid.
-    latest_tile_sets = await db.all(latest_raster_tile_sets, {"grid": grid})
+    sql = _get_data_environment_sql(version_overrides)
+    latest_tile_sets = await db.all(db.text(sql), {"grid": grid})
```
danscales, line 837: Would be nice to change 'latest_tile_sets' to 'matching_tile_sets', since they are no longer just latest tile versions.

danscales (approved):
Looks good, except for the one case of wrong args for _query_raster() call.  Other comments are minor.

Nice to be able to specify versions other than the latest!

jterry64:
@danscales thanks for the comments, resolved them and will merge the PR at the end of the day if you're fine with my resolutions

PR #636, opened 2025-02-25
GTC-2708 Fix deleting of the tile cache when deleting a version

danscales:
GTC-2708 Fix deleting of the tile cache when deleting a version

The main issue was that the correct lifecycle action permission is actually s3:PutLifecycleConfiguration, but we've always had s3:PutBucketLifecycleConfiguration in the policy templates.

A second issue is that the expiration time for a bucket/folder must be a date specified as midnight UTC. We were just using the current time, so that caused an error when trying to set the lifecycle rule.

In both cases, we weren't seeing the error, because delete_all_assets was called from a background task, so any exception was getting lost without being logged. So, I added a try-except around the call to put_bucket_lifecycle_configuration(), so we log future errors (and don't kill the task).

Given these fixes, I was able to re-enable the use of expire_s3_objects() and it worked!

### DIFF
--- a/app/tasks/aws_tasks.py
+++ b/app/tasks/aws_tasks.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional
 
 from botocore.exceptions import ClientError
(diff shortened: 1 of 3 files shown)


app/tasks/aws_tasks.py
```diff
@@ -121,10 +125,17 @@ def _update_lifecycle_rule(bucket, rule) -> Dict[str, Any]:
...
+    logger.info(f"Add lifecycle configuration rule {rules} to bucket {bucket}")
+    response = {}
+    try:
+        response = client.put_bucket_lifecycle_configuration(
+            Bucket=bucket, LifecycleConfiguration={"Rules": rules}
+        )
+    except Exception as e:
+        # Make sure to print out any exception, since this function is probably
+        # running in a background task, so any thrown exception may get lost.
+        logger.info(f"lifecycle exception {str(e)}, {e.__class__.__name__}, {e.args}")
```
jterry64, line 137: I think logger.error or logger.warning would be more appropriate here

jterry64 (approved):

dmannarino (approved):
Hot diggity! Thanks!

solomon-negusse (approved):

PR #644, opened 2025-03-03
GTC-3091: Get GADM 4.1 features from Data API, simplify if requested

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- There are geostore endpoints which forward requests to the RW API to get GADM 3.6 features.

Issue Number: GTC-3091

## What is the new behavior?

- For the endpoints /admin and /admin/list endpoints, make it possible (via query params) to get 4.1 features from the Data API as well.
- Add the ability to specify simplification.

- [ ] Yes
- [x] No


app/crud/geostore.py
```diff
@@ -115,3 +132,253 @@ async def create_user_area(geometry: Geometry) -> Geostore:
...
+        db.text("adm_level=:adm_level").bindparams(adm_level=str(adm_level))
+    ]
+
+    # gid_0 is just a three-character value, but all more specific ids are
+    # followed by an underscore (which has to be escaped because normally in
+    # SQL an underscore is a wildcard) and a revision number (for which we
+    # use an UN-escaped underscore).
+    level_id_pattern: str = country_id
+
+    if adm_level == 0:  # Special-case to avoid slow LIKE
```
jterry64, line 243: not part of this PR per se, but should we make sure to add the B-tree soon so LIKE queries are fast?

app/crud/geostore.py
```diff
@@ -115,3 +132,253 @@ async def create_user_area(geometry: Geometry) -> Geostore:
...
+            level_id_pattern = ".".join((level_id_pattern, subregion_id))
+        level_id_pattern += r"\__"
+
+        # Adjust for any errata
+        level_id_pattern = fix_id_pattern(
+            adm_level, level_id_pattern, admin_provider, admin_version
+        )
+
+        where_clauses.append(
+            db.text(f"gid_{adm_level} LIKE :level_id_pattern").bindparams(
```
jterry64, line 263: Shouldn't the ID be deterministic, since they have to explicitly pass the iso/adm1/adm2? In that case, why use `LIKE` instead of direct `=`? That would definitely run fastest

app/crud/geostore.py
```diff
@@ -115,3 +132,253 @@ async def create_user_area(geometry: Geometry) -> Geostore:
...
+    return rows
+
+
+async def get_first_row(sql: Select):
+    row = await db.first(sql)
+
+    return row
+
+
+async def get_gadm_geostore(
```
jterry64, line 193: You mentioned you might be working on this, but either now or future PR, might be good to break this up into a few subfunctions to make it more readable (e.g. `add_simplify`, `add_admin_id_filter`, etc.

app/crud/geostore.py
```diff
@@ -115,3 +132,253 @@ async def create_user_area(geometry: Geometry) -> Geostore:
...
+            **info.dict(),
+            id1=int(extract_level_id(1, level_id)),
+        )
+    if adm_level == 2:
+        info = Adm2BoundaryInfo(
+            **info.dict(),
+            id2=int(extract_level_id(2, level_id)),
+        )
+
+    return AdminGeostore.parse_obj(
```
jterry64, line 360: Also can be done later, but I feel like it'd be more elegant to have AdminGeostore have a constructor to just take these params, and then have a function that outputs them in this format. But maybe just me prefernce.

app/utils/gadm.py
```diff
@@ -0,0 +1,37 @@
+GADM_41_IDS_MISSING_REVISION = (
+    "IDN.35.4",
+    "IDN.35.8",
+    "IDN.35.9",
+    "IDN.35.13",
+    "IDN.35.14",
+)
+
+
+def extract_level_id(adm_level: int, id_string: str):
```
jterry64, line 10: A note for next time we update GADM, if at all: we should have a script that normalizes all IDs and splits them into separate columns, instead of having to code around all these things

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -178,7 +184,95 @@
 
 
 @pytest.mark.asyncio
-async def test_get_admin_geostore(async_client: AsyncClient, monkeypatch: MonkeyPatch):
+async def test_get_admin_country_geostore_gfw_branch(
+    async_client: AsyncClient, monkeypatch: MonkeyPatch
+):
+    # CursorResult = MagicMock(cursor)
```
jterry64, line 190: comment to delete? or is it for future debugging?

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -178,7 +184,95 @@
...
+    # mock_session.execute.return_value = [
+    #     CursorResult(tup1),
+    #     CursorResult(tup2)
+    # ]
+    # CursorResult.return_value._asdict.side_effect = [tup1, tup2]
+
+    mock_get_first_row: AsyncMock = AsyncMock(return_value=None)
+
+    monkeypatch.setattr(crud_geostore, "get_first_row", mock_get_first_row)
```
jterry64, line 199: Why are you mixing monkeypatch and mock library instead of just using the mock patch method?

jterry64 (changes requested):
Looks good, I had a couple of style comment, but if pressed for time the only comment I really need answered is about why we're using `LIKE` in the query.

gtempus (approved):
Nice work pulling this together. I'm glad it worked not to store these simplified geometries. 💪 

I only have additional testing suggestions you may or may not want to entertain.

app/crud/geostore.py
dmannarino: Even if the adm0/adm1/adm2 combination were unique (probably, but we haven't checked), the the revisions are the fly in the ointment. :( Every identifier lower than adm0 has an underscore and a revision on the end, and there's no way to know what the revision will be. Most are 1 or 2, but it could be anything. So we have to match on `_<something>` :(

app/crud/geostore.py
dmannarino: Yeah, it's really big. Will factor into some smaller functions in an isolated commit.

tests_v2/unit/app/routes/geostore/test_geostore.py
dmannarino: Yep, will get rid of it.

tests_v2/unit/app/routes/geostore/test_geostore.py
dmannarino: A good question. Will switch to 100% mock.

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -178,7 +184,95 @@
...
+        "gid_1 AS level_id, name_1 AS name, ST_AsGeoJSON(geom) AS geojson "
+        '\nFROM gadm_administrative_boundaries."v4.1.64" \n'
+        r"WHERE adm_level='1' AND gid_1 LIKE 'MEX.1\__'"
+    )
+    actual_sql = str(
+        mock_get_first_row.call_args.args[0].compile(
+            compile_kwargs={"literal_binds": True}
+        )
+    )
+    assert actual_sql == expected_sql
```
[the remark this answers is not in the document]
dmannarino: Good point, thanks! Will narrow the gap between where I call and where I test.

app/crud/geostore.py
dmannarino: I imagine we should, though I'm not 100% sure a B-tree index is the thing to do. I see suggestions that it can work if we are matching a prefix (a beginning of a pattern), which we are, but it sounds like maybe only if we use the C locale? I'm a little unsure, and also unsure of the ramifications of that (but suspect it won't play nice with our Unicode strings). And then there are these mentions of a trigram index. Here's where I was reading: https://dba.stackexchange.com/a/10696

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -316,3 +410,60 @@ async def test_get_geostore_gfw_branch(
 
     assert mock_proxy_get_geostore.called is False
     assert mock_get_gfw_geostore_from_any_dataset.called is True
+
+
+@pytest.mark.asyncio
+async def test_get_admin_list_rw_branch(
```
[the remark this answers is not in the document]
dmannarino: Roger! Will rename.

app/crud/geostore.py
danscales: This query shows the only gid_2's that don't end in _1 or _2 - which are only the 5 IDN exceptions and "?".  So I would definitely recommend that we just match against _1 and _2 (after dealing with the exceptions), so we're not doing the slow LIKE queries.  We could file a separate to fix this ticket, so we don't slow down this PR.

https://data-api.globalforestwatch.org/dataset/gadm_administrative_boundaries/v4.1.64/query/json?sql=SELECT%20%20gid_2%20FROM%20data%20WHERE%20gid_2%20NOT%20LIKE%20%27%25_1%27%20AND%20gid_2%20NOT%20LIKE%20%27%25_2%27&&x-api-key=3dc1a39a-60b2-4352-bb92-1413257dafd9

(sql is just: SELECT  gid_2 FROM data WHERE gid_2 NOT LIKE '%_1' AND gid_2 NOT LIKE '%_2'")

dmannarino: I'll create a ticket, thanks for checking!

app/utils/gadm.py
dmannarino: I have a script which generates the normalized (de-accented, de-capitalized) fields, but didn't know about these anomalies at the time.

dmannarino: I will modify it to take care of these anomalies, in case we have to import again anyway. We could even have it strip the _<revision>, and do away with the LIKE!

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -434,7 +427,7 @@ async def test_get_admin_list_rw_branch_36(
     async_client: AsyncClient, monkeypatch: MonkeyPatch
 ):
     url = "/geostore/admin/list"
-    params = {"adminVersion": "3.6"}
+    params = {"source[version]": "3.6"}
```
gtempus, line 427: Good catch!

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -423,7 +423,7 @@ async def test_get_admin_list_rw_branch(
 
 
 @pytest.mark.asyncio
-async def test_get_admin_list_rw_branch_36(
+async def test_get_admin_list_proxies_to_rw_when_gadm_36_requested(
```
gtempus, line 423: Naming is a lot better. 💯

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -442,7 +442,7 @@ async def test_get_admin_list_rw_branch_36(
 
 
 @pytest.mark.asyncio
-async def test_get_admin_list_gfw_branch_41(
+async def test_get_admin_list_gets_gadm_41_from_data_api(
```
gtempus: Same here. 💯

tests_v2/unit/app/routes/geostore/test_geostore.py
```diff
@@ -406,7 +406,7 @@ async def test_get_geostore_gfw_branch(
 
 
 @pytest.mark.asyncio
-async def test_get_admin_list_rw_branch(
+async def test_get_admin_list_proxies_to_rw_when_no_source_info_is_provided(
```
gtempus, line 408: ditto 💯

gtempus (approved):
Love the re-org of the tests and the names. Our future selves thank you, @dmannarino! 🙇

app/crud/geostore.py
dmannarino: https://gfw.atlassian.net/browse/GTC-3183

app/crud/geostore.py
dmannarino: Thanks for creating a ticket for this, @danscales !

app/crud/geostore.py
dmannarino: Hmm. I think I prefer it this way because it's how we tend to do it elsewhere and I'd like to be consistent. But that is an interesting idea.

PR #646, opened 2025-03-05
Various fixes for drivers endpiints

jterry64:
Uses hyphen in name of path, broken for if default versions of datasets don't exist in environment.

Issue Number: GTC-3167

## What is the new behavior?

- I swear I changed all these paths in the previous PR, but somehow got lost. But snake case for everything, including API path, to be consistent.
- Missed that the call in the compute task to _query_dataset_json always used the default version for umd_tree_cover_loss as the default asset, but should use the version override if it exists.
- While trying to find this, ending up adding other unit tests for queries module to make sure everything is covered.

### DIFF
--- a/tests_v2/unit/app/routes/datamart/test_land.py
+++ b/tests_v2/unit/app/routes/datamart/test_land.py
@@ -31,13 +31,13 @@ async def test_get_tree_cover_loss_by_drivers_not_found(
         params = {"x-api-key": api_key, "geostore_id": geostore, "canopy_cover": 30}
 
         response = await async_client.get(
-            "/v0/land/tree-cover-loss-by-driver", headers=headers, params=params
+            "/v0/land/tree_cover_loss_by_driver", headers=headers, params=params
         )
 
         assert response.status_code == 404
 
         resource_id = _get_resource_id(
-            "tree-cover-loss-by-driver", geostore, 30, DEFAULT_LAND_DATASET_VERSIONS
+            "tree_cover_loss_by_driver", geostore, 30, DEFAULT_LAND_DATASET_VERSIONS
         )
         mock_get_resources.assert_awaited_with(resource_id)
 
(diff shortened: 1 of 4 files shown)


tests_v2/unit/app/routes/datamart/test_land.py
```diff
@@ -103,14 +103,14 @@ async def test_get_tree_cover_loss_by_drivers_with_overrides(
         )
 
         response = await async_client.get(
-            "/v0/land/tree-cover-loss-by-driver?x-api-key={api_key}&geostore_id={geostore_id}&canopy_cover=30&dataset=umd_tree_cover_loss&version=v1.8&dataset=umd_tree_cover_density_2000&version=v1.6",
+            "/v0/land/tree_cover_loss_by_driver?x-api-key={api_key}&geostore_id={geostore_id}&canopy_cover=30&dataset=umd_tree_cover_loss&version=v1.8&dataset=umd_tree_cover_density_2000&version=v1.6",
```
dmannarino, line 106: Is this supposed to be an f-string? I see `{api_key}` and `{geostore_id}` but no "f"

tests_v2/unit/app/routes/datasets/test_query.py
```diff
@@ -823,6 +838,120 @@ async def test_query_batch_uri(
...
+        assert args[0][5] == {dataset_name: "v2"}
+        assert result == {"data": []}
+
+
+@pytest.mark.asyncio
+async def test_query_raster_lambda_with_overrides(
+    generic_raster_version,
+    geostore,
+):
+    dataset_name, _ = generic_raster_version
```
dmannarino, line 895: There are actually three things in here, hence this failure: https://github.com/wri/gfw-data-api/actions/runs/13687801477/job/38275023712#step:3:5005

tests_v2/unit/app/routes/datamart/test_land.py
jterry64: That's a good question... not sure how it's passing!

tests_v2/unit/app/routes/datasets/test_query.py
jterry64: Ah, I had accidentally removed that last minute due to flake8 message, fixed!

tests_v2/unit/app/routes/datamart/test_land.py
jterry64: @dmannarino fixed, but it was passing because API key was already sent in the header, and the place geostore_id was being used was mocked. So I think the test is actually testing things.

danscales (approved):

PR #648, opened 2025-03-06
Various fixes for drivers endpoints

jterry64:
- Snake case for everything, including API path, to be consistent. 
-Missed that the call in the compute task to query_dataset_json always used the default version for 
- umd_tree_cover_loss as the default asset, but should use the version override if it exists. While trying to find this, ending up adding other unit tests for queries module to make sure everything is covered.


gtempus (approved):
Nice test additions, @jterry64 💪

PR #649, opened 2025-03-07
GTC-3163: Fix providing API keys for new GADM geostore endpoints; send ours to RW

dmannarino:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Providing a Data API key with one of the new RW geostore endpoints yields a 404.

Issue Number: GTC-3163

## What is the new behavior?

- Providing an API key is optional (for now), but works (the Data API's RW API key is used in the request to the RW API)

- [ ] Yes
- [x] No


gtempus (commented):
If these changes mean that the API key is required, won't that break Flagship if this goes to production?

If so, shouldn't they start sending their API key first and we ensure that we don't respond with an error code?

dmannarino:
> If these changes mean that the API key is required, won't that break Flagship if this goes to production?
> 
> If so, shouldn't they start sending their API key first and we ensure that we don't respond with an error code?

Blast, you're right. We have to use a phased approach. :( Will fix.

app/settings/globals.py
```diff
@@ -146,6 +146,9 @@
 AWS_SECRETSMANAGER_URL = config("AWS_SECRETSMANAGER_URL", cast=str, default=None)
 
 RW_API_URL = config("RW_API_URL", cast=str, default=None)
+RW_API_KEY = json.loads(
+    config("RW_API_KEY", cast=str, default=json.dumps({"api_key": "foo"}))
```
solomon-negusse, line 150: setting a default with an invalid key will cause a 404.. maybe  we don't send `x-api-key` header  to RW API if it's not passed?

solomon-negusse (approved):
Looks good, just one suggested fix that's not pressing

jterry64 (approved):

app/settings/globals.py
dmannarino: It's true that this is a bad default, but I don't intend to ever let this variable go un-set in a deployed environment. I chose this to satisfy type checking, though I'm not happy about it. Somehow I seem to have forgotten I don't NEED to provide a default, and it will complain on startup if it's not populated. I will remove the default. 
WRT your suggestion, now that we are sending the GFW Data API's RW API key the presence of the header in the client's request is immaterial. I think it's better to always send the API key because it gives us information about the usage.

solomon-negusse: My message was a bit confusing - what I meant was in environments where it's not set. I saw that `RW_API_KEY` wasn't added to docker-compose-dev file and thought a simpler solution when using these endpoints locally was to add the header optionally rather than needing to add the env variable to the file.

dmannarino: Okay, I've thought a lot about this. I think I'll have it send the key if it is defined in the environement. It always will be in deployed environments, and it won't break locally unless you explicitly provide a bad one. Maybe that's what you meant, but I was thinking you meant only send it if the client sends a GFW one. :)

PR #652, opened 2025-03-10
Dict dataset_version params

jterry64:
For the first iteration of pinning datasets and versions, I couldn't figure out how to send a dictionary as a FastAPI parameter, so I made two list parameters, `dataset` and `version`. The code just zipped these together in the order provided, which could lead to all kinds of mistakes.

I figured out how we can send a dictionary parameter like `dataset_version[dataset]=version` by using the `Request` object in FastAPI. This is definitely more ideal and what we originally agreed on.


gtempus (approved):
I like this implementation much better. The dataset and the version are clearly associated now. 💯

solomon-negusse (approved):

PR #656, opened 2025-03-12
GTC-3172: Support GADM areas in drivers endpoints

gtempus:
## Pull request checklist

## Pull request type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Administrative boundary areas are not supported. The client must have a geostore ID

Issue Number: GTC-3172

## What is the new behavior?
Clients can now use Administrative Boundaries with requests.

GET (query parameters):
`aoi[type]=admin&aoi[country]=BRA&aoi[region]=12&aoi[subregion]=23`

POST (body):
```json
{
  "aoi": {
    "type": "admin",
    "country": "BRA",
    "region": "12",
    "subregion": "23"
  }
}
```
### NOTE
There is now a grouping parameter (`aoi`) and a discriminator (`aoi[type]`)
The discriminator makes for very nice openapi documentation and allows much richer validation for pydantic unions
https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions

**Only GADM 4.1 is currently supported**

- [x] Yes
- [ ] No

### DIFF
--- a/app/crud/geostore.py
+++ b/app/crud/geostore.py
@@ -9,7 +9,6 @@ from sqlalchemy.sql import Select, label
 from sqlalchemy.sql.elements import Label, TextClause
 
 from app.application import db
-from app.crud.tasks import create_or_update_task
 from app.errors import (
     BadAdminSourceException,
     BadAdminVersionException,
@@ -213,7 +212,7 @@ async def get_gadm_geostore_id(
         src_table,
         subregion_id,
     )
-    return await row.gfw_geostore_id
+    return row.gfw_geostore_id
 
 
 async def build_gadm_geostore(
--- a/app/crud/geostore.py
+++ b/app/crud/geostore.py
@@ -200,13 +200,7 @@ async def build_gadm_geostore(
     region_id: str | None = None,
     subregion_id: str | None = None,
 ) -> AdminGeostore:
-    dv: Tuple[str, str] = await admin_params_to_dataset_version(
-        admin_provider, admin_version
-    )
-    dataset, version = dv
-
-    src_table: Table = db.table(version)
-    src_table.schema = dataset
+    src_table = await get_versioned_dataset(admin_provider, admin_version)
 
     columns_etc: List[Column | Label] = [
         db.column("adm_level"),
@@ -231,16 +225,43 @@ async def build_gadm_geostore(
             )
         )
 
+    sql: Select = db.select(columns_etc).select_from(src_table)
+
+    sql = await add_where_clauses(adm_level, admin_provider, admin_version, country_id, region_id, sql, subregion_id)
+
+    row = await get_first_row(sql)
+    if row is None:
+        raise RecordNotFoundError(
+            f"Admin boundary not found in {admin_provider} version {admin_version}"
+        )
+
+    if row.geojson is None:
+        raise GeometryIsNullError(
+            "GeoJSON is None, try reducing or eliminating simplification."
+        )
+
+    return await form_admin_geostore(
+        adm_level=adm_level,
+        admin_version=admin_version,
+        area=float(row.gfw_area__ha),
+        bbox=[float(val) for val in row.gfw_bbox],
+        name=str(row.name),
+        geojson=json.loads(row.geojson),
+        geostore_id=str(row.gfw_geostore_id),
+        level_id=str(row.level_id),
+        simplify=simplify,
+    )
+
+
+async def add_where_clauses(adm_level, admin_provider, admin_version, country_id, region_id, sql, subregion_id):
     where_clauses: List[TextClause] = [
         db.text("adm_level=:adm_level").bindparams(adm_level=str(adm_level))
     ]
-
     # gid_0 is just a three-character value, but all more specific ids are
     # followed by an underscore (which has to be escaped because normally in
     # SQL an underscore is a wildcard) and a revision number (for which we
     # use an UN-escaped underscore).
     level_id_pattern: str = country_id
-
     if adm_level == 0:  # Special-case to avoid slow LIKE
         where_clauses.append(
             db.text("gid_0=:level_id_pattern").bindparams(
@@ -265,34 +286,19 @@ async def build_gadm_geostore(
                 level_id_pattern=level_id_pattern
             )
         )
-
-    sql: Select = db.select(columns_etc).select_from(src_table)
-
     for clause in where_clauses:
         sql = sql.where(clause)
+    return sql
 
-    row = await get_first_row(sql)
-    if row is None:
-        raise RecordNotFoundError(
-            f"Admin boundary not found in {admin_provider} version {admin_version}"
-        )
 
-    if row.geojson is None:
-        raise GeometryIsNullError(
-            "GeoJSON is None, try reducing or eliminating simplification."
-        )
-
-    return await form_admin_geostore(
-        adm_level=adm_level,
-        admin_version=admin_version,
-        area=float(row.gfw_area__ha),
-        bbox=[float(val) for val in row.gfw_bbox],
-        name=str(row.name),
-        geojson=json.loads(row.geojson),
-        geostore_id=str(row.gfw_geostore_id),
-        level_id=str(row.level_id),
-        simplify=simplify,
+async def get_versioned_dataset(admin_provider, admin_version):
+    dv: Tuple[str, str] = await admin_params_to_dataset_version(
+        admin_provider, admin_version
     )
+    dataset, version = dv
+    src_table: Table = db.table(version)
+    src_table.schema = dataset
+    return src_table
 
 
 async def get_gadm_geostore(
--- a/app/crud/geostore.py
+++ b/app/crud/geostore.py
@@ -9,6 +9,7 @@ from sqlalchemy.sql import Select, label
 from sqlalchemy.sql.elements import Label, TextClause
 
 from app.application import db
+from app.crud.tasks import create_or_update_task
 from app.errors import (
     BadAdminSourceException,
     BadAdminVersionException,
@@ -190,7 +191,7 @@ async def get_first_row(sql: Select):
     return row
 
 
-async def get_gadm_geostore(
+async def build_gadm_geostore(
     admin_provider: str,
     admin_version: str,
     adm_level: int,
@@ -198,7 +199,7 @@ async def get_gadm_geostore(
     country_id: str,
     region_id: str | None = None,
     subregion_id: str | None = None,
-) -> AdminGeostoreResponse:
+) -> AdminGeostore:
     dv: Tuple[str, str] = await admin_params_to_dataset_version(
         admin_provider, admin_version
     )
@@ -281,7 +282,7 @@ async def get_gadm_geostore(
             "GeoJSON is None, try reducing or eliminating simplification."
         )
 
-    geostore: AdminGeostore = await form_admin_geostore(
+    return await form_admin_geostore(
         adm_level=adm_level,
         admin_version=admin_version,
         area=float(row.gfw_area__ha),
@@ -291,6 +292,26 @@ async def get_gadm_geostore(
         geostore_id=str(row.gfw_geostore_id),
         level_id=str(row.level_id),
         simplify=simplify,
+    )
+
+
+async def get_gadm_geostore(
+    admin_provider: str,
+    admin_version: str,
+    adm_level: int,
+    simplify: float | None,
+    country_id: str,
+    region_id: str | None = None,
+    subregion_id: str | None = None,
+) -> AdminGeostoreResponse:
+    geostore: AdminGeostore = await build_gadm_geostore(
+        admin_provider=admin_provider,
+        admin_version=admin_version,
+        adm_level=adm_level,
+        simplify=simplify,
+    country_id=country_id,
+        region_id=region_id,
+        subregion_id=subregion_id,
     )
 
     return AdminGeostoreResponse(data=geostore)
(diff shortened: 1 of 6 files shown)


app/routes/datamart/land.py
```diff
@@ -41,28 +45,107 @@
 
 router = APIRouter()
 
+def _parse_dataset_versions(request: Request) -> Dict[str, str]:
```
jterry64, line 48: just curious, why'd you move these functions to the top? is it more standard to have private function first?

app/routes/datamart/land.py
```diff
@@ -41,28 +45,107 @@
...
+                )
+            }
+        ]
+    },
 )
 async def tree_cover_loss_by_driver_search(
     *,
-    request: Request,
-    geostore_id: UUID = Query(..., title="Geostore ID"),
+    aoi: AreaOfInterest = Depends(_parse_area_of_interest),
```
jterry64, line 140: I feel like `aoi` is a very acronym internally, but I'm not sure about externally. Should we make it more explicit, e.g. `area_of_interest`? Or do you think it's fine as long as it's clearly used as terminology in the docs?

app/routes/datamart/land.py
```diff
@@ -141,7 +225,7 @@ async def tree_cover_loss_by_driver_post(
     background_tasks.add_task(
```
jterry64, line 225: I think you might be missing one big part, which is actually modifying the task to deal with GADM areas per the method discussed here: https://gfw.atlassian.net/browse/GTC-3116

Basically, the bounds for an admin area might be too large, and we don't want to pass simplify boundaries to keep accuracy. So we need to pull simplified boundaries, buffered boundaries from the database, and then pass them to OTF with a filter using the rasterized GADM.

We can potentially merge this as is, and then add that optimization afterwards, although some countries with complicated bounds might not work (they'll hit a lambda limit).

jterry64 (changes requested):

app/routes/datamart/land.py
jterry64: Or alternatively, pass a simplification parameter when getting the geostore just so it'll always work, but then we'll need to do some invalidation later once we have the full method since results might not be accurate.

app/routes/datamart/land.py
gtempus: Good question; it's because of how Python evaluates default values for formal parameters in function definitions.

Python evaluates a default value only once when the module is loaded. In this concrete example, `Depends` is called when the default is evaluated. At that point, `_parse_area_of_interest` and `_parse_dataset_versions` need to be defined. In the previous version, they were lower down in the module and not defined yet.

app/routes/datamart/land.py
gtempus: I think we gain more readability by keeping it short. So, I'd vote for a definition in the docs.

That said, I hear my kids and their friends saying `AOI` or `POI` all the time when gaming. 😉

app/routes/datamart/land.py
gtempus: 🤦 I remember us talking about this, @jterry64. 
Can we meet up after today's Drivers/Datamart huddle to revisit what an implementation might look like?
I remember suggesting a new implementation of `_compute_tree_cover_loss_by_driver`

dmannarino: I'm interested in the solution to this as well. Mind if I join in?

dmannarino:
In general this looks good! Holding off approving due to the simplification question. Also I think there's a little linting that went awry (either that or my knowledge of standards is out of date). Would you mind trying to run the pre-commit hooks on the involved files again?

app/crud/geostore.py
```diff
@@ -230,16 +246,43 @@ async def get_gadm_geostore(
             )
         )
```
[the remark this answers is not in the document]
gtempus: Good idea! Thank you!

app/models/pydantic/datamart.py
```diff
@@ -1,12 +1,68 @@
...
+
+    @root_validator
+    def check_region_subregion(cls, values):
+        region = values.get("region")
+        subregion = values.get("subregion")
+        if subregion is not None and region is None:
+            raise ValueError("region must be specified if subregion is provided")
+        return values
+
+    @validator('provider', pre=True, always=True)
```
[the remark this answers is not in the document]
gtempus: Excellent observation. That's what I thought, too!
However, when testing compatibility with the Swagger UI, it seems that Swagger likes to add the missing properties and populate them with `None`. 🙄 As a result, pydantic freaks out with validation errors.

app/routes/datamart/land.py
```diff
@@ -41,28 +45,107 @@
...
+        raise HTTPException(status_code=422, detail=e.errors())
+
 
 @router.get(
     "/tree_cover_loss_by_driver",
     response_class=ORJSONResponse,
     response_model=DataMartResourceLinkResponse,
     tags=["Land"],
     status_code=200,
+    openapi_extra={
```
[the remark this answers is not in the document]
gtempus: Good idea. I was thinking about that, too. I could put it in `datamart/__init__.py`?

tests_v2/unit/app/routes/datamart/test_land.py
```diff
@@ -140,10 +140,10 @@ async def test_get_tree_cover_loss_by_drivers_with_malformed_overrides(
     origin = payload["domains"][0]
 
     headers = {"origin": origin}
-    params = {"x-api-key": api_key, "geostore_id": geostore, "canopy_cover": 30}
+    params = {"x-api-key": api_key, "aoi[type]": "geostore", "aoi[geostore_id]": geostore, "canopy_cover": 30}
 
     response = await async_client.get(
-        "/v0/land/tree_cover_loss_by_driver?x-api-key={api_key}&geostore_id={geostore_id}&canopy_cover=30&dataset_version[umd_tree_cover_loss]]=v1.8&dataset_version[umd_tree_cover_density_2000]=v1.6",
+        f"/v0/land/tree_cover_loss_by_driver?dataset_version[umd_tree_cover_loss]]=v1.8&dataset_version[umd_tree_cover_density_2000]=v1.6",
```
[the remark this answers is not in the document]
gtempus: Good spot! But, that's actually what the test is looking for. 😄 
`test_get_tree_cover_loss_by_drivers_with_malformed_overrides`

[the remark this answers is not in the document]
solomon-negusse: Lol smh (at myself)

app/routes/datamart/land.py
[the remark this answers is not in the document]
solomon-negusse: I think that's good place for it. BTW, I added data mart to the API docs locally and the dropdown UI for AOIs looks neat.

app/models/pydantic/datamart.py
```diff
@@ -1,12 +1,68 @@
...
+
+
+class AreaOfInterest(StrictBaseModel, ABC):
+    @abstractmethod
+    async def get_geostore_id(self) -> UUID:
+        """Return the unique identifier for the area of interest."""
+        pass
+
+
+class GeostoreAreaOfInterest(AreaOfInterest):
```
solomon-negusse, line 21: One last comment:   I feel `CustomAreaOfInterest` or `UserAreaOfInterest` may be more fitting name for the public facing doc

app/routes/datamart/land.py
gtempus: We decided to break this functionality into its own Jira ticket: https://gfw.atlassian.net/browse/GTC-3219

app/routes/datamart/land.py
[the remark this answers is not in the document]
gtempus: refactored in 21e1cdffe07183694badb1c8c2d5efa086c37271

app/crud/geostore.py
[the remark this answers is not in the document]
gtempus: refactored in 552d62338c2b6b6c62db06d0c226b03057699dde

app/models/pydantic/datamart.py
gtempus: I personally feel that those are too vague. But I'd like to hear what other folks think.
CC @jterry64 @dmannarino

gtempus:
@dmannarino, I ran the pre-commit hook in this commit: 6193fc5140baa2b6e374e1b233a8409c09dbcca2

app/crud/geostore.py
```diff
@@ -9,6 +9,7 @@
 from sqlalchemy.sql.elements import Label, TextClause
 
 from app.application import db
+from app.crud.tasks import create_or_update_task
```
danscales, line 12: This new import is not actually used. maybe remove?

app/crud/geostore.py
```diff
@@ -190,22 +191,41 @@ async def get_first_row(sql: Select):
...
+        admin_provider,
+        admin_version,
+        columns_etc,
+        country_id,
+        region_id,
+        src_table,
+        subregion_id,
     )
-    dataset, version = dv
+    return await row.gfw_geostore_id
```
danscales, line 216: I think the 'await' keyword is unnecessary, right? If so, maybe remove to avoid confusion.

app/models/pydantic/datamart.py
```diff
@@ -43,7 +104,9 @@ class DataMartResourceLinkResponse(Response):
 
 
 class TreeCoverLossByDriverIn(StrictBaseModel):
-    geostore_id: UUID
+    aoi: Union[GeostoreAreaOfInterest, AdminAreaOfInterest] = Field(
```
danscales, line 107: Not your change, but line 95 above, the type should be Optional[DataMartMetaDdata], so it can have a value of None.

tests_v2/unit/app/crud/test_geostore.py
```diff
@@ -117,3 +116,107 @@ async def test_get_gadm_geostore_generates_correct_sql_for_subregion_lookup(
 
     assert mock_get_first_row.called is True
     assert actual_sql == expected_sql
+
+
+class TestGadmGeostoreIDLookup:
```
danscales, line 121: **Just wondering what this class means or is useful for here?

Since the methods below are inside a class, shouldn't they have a self argument. I guess it may just work, because async_client is just assumed to be their self argument, but is never actually used.

tests_v2/unit/app/routes/datamart/test_land.py
```diff
@@ -140,10 +155,15 @@ async def test_get_tree_cover_loss_by_drivers_with_malformed_overrides(
...
+    params = {
+        "x-api-key": api_key,
+        "aoi[type]": "geostore",
+        "aoi[geostore_id]": geostore,
+        "canopy_cover": 30,
+    }
 
     response = await async_client.get(
-        "/v0/land/tree_cover_loss_by_driver?x-api-key={api_key}&geostore_id={geostore_id}&canopy_cover=30&dataset_version[umd_tree_cover_loss]]=v1.8&dataset_version[umd_tree_cover_density_2000]=v1.6",
+        f"/v0/land/tree_cover_loss_by_driver?dataset_version[umd_tree_cover_loss]]=v1.8&dataset_version[umd_tree_cover_density_2000]=v1.6",
```
danscales, line 166: Don't need the f here anymore, but did you put it in for consistency, or did you mean to use a variable reference somewhere in the string?

danscales (commented):
Just a number of minor comments. Looks good.

app/crud/geostore.py
gtempus: Good catch! Thank you, @danscales

app/crud/geostore.py
gtempus: Yup! Not sure how that `await` got there. 😄

tests_v2/unit/app/crud/test_geostore.py
gtempus: You're right about `self`. It was a cut-and-paste error. 🤦 
I'm using the class as [an organizational feature](https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class). It sets off the tests nicely in the test results output.

tests_v2/unit/app/routes/datamart/test_land.py
gtempus: Good catch. Thank you!

app/models/pydantic/datamart.py
gtempus: addressed in 7cd6d1b8a5e3795bf2b582e4e14241a5caa7eea6

tests_v2/unit/app/crud/test_geostore.py
gtempus: addressed in 7cd6d1b8a5e3795bf2b582e4e14241a5caa7eea6

app/crud/geostore.py
gtempus: addressed in 7cd6d1b8a5e3795bf2b582e4e14241a5caa7eea6

app/crud/geostore.py
gtempus: addressed in 7cd6d1b8a5e3795bf2b582e4e14241a5caa7eea6

tests_v2/unit/app/routes/datamart/test_land.py
gtempus: addressed in 7cd6d1b8a5e3795bf2b582e4e14241a5caa7eea6

app/models/pydantic/datamart.py
```diff
@@ -1,12 +1,68 @@
...
-from pydantic import Field
+from pydantic import Field, root_validator, validator
 
 from app.models.pydantic.responses import Response
 
 from .base import StrictBaseModel
+from ...crud.geostore import get_gadm_geostore_id
+
+
+class AreaOfInterest(StrictBaseModel, ABC):
```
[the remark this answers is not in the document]
gtempus: I created a Jira card for this: https://gfw.atlassian.net/browse/GTC-3227

app/models/pydantic/datamart.py
gtempus: We're going to keep the current naming for now. Will revisit in the future.

PR #658, opened 2025-03-21
Gtc 3187/datamart global analysis

solomon-negusse:
This adds to TCL by drivers data mart endpoint a way to serve pre-computed global results.
This is based on Gary's data mart admin AOI support PR and will be updated accordingly.

### DIFF
--- a/app/crud/geostore.py
+++ b/app/crud/geostore.py
@@ -9,7 +9,6 @@ from sqlalchemy.sql import Select, label
 from sqlalchemy.sql.elements import Label, TextClause
 
 from app.application import db
-from app.crud.tasks import create_or_update_task
 from app.errors import (
     BadAdminSourceException,
     BadAdminVersionException,
--- a/app/crud/geostore.py
+++ b/app/crud/geostore.py
@@ -200,13 +200,7 @@ async def build_gadm_geostore(
     region_id: str | None = None,
     subregion_id: str | None = None,
 ) -> AdminGeostore:
-    dv: Tuple[str, str] = await admin_params_to_dataset_version(
-        admin_provider, admin_version
-    )
-    dataset, version = dv
-
-    src_table: Table = db.table(version)
-    src_table.schema = dataset
+    src_table = await get_versioned_dataset(admin_provider, admin_version)
 
     columns_etc: List[Column | Label] = [
         db.column("adm_level"),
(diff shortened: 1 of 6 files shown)


app/crud/geostore.py
```diff
@@ -264,33 +307,38 @@ async def get_gadm_geostore(
...
         admin_version=admin_version,
-        area=float(row.gfw_area__ha),
-        bbox=[float(val) for val in row.gfw_bbox],
-        name=str(row.name),
-        geojson=json.loads(row.geojson),
-        geostore_id=str(row.gfw_geostore_id),
-        level_id=str(row.level_id),
+        adm_level=adm_level,
         simplify=simplify,
+    country_id=country_id,
```
danscales, line 339: Weird indentation, maybe fix?

app/models/pydantic/datamart.py
```diff
@@ -1,12 +1,68 @@
...
+from pydantic import Field, root_validator, validator
 
 from app.models.pydantic.responses import Response
 
 from .base import StrictBaseModel
+from ...crud.geostore import get_gadm_geostore_id
+
+
+class AreaOfInterest(StrictBaseModel):
+    async def get_geostore_id(self) -> UUID:
```
danscales, line 14: Maybe add '@abstractmethod' here (from abc package), since this has to be overridden (since not returning anything).

Or maybe throw an exception if this method is actually called? I see that Global() does not actually define a get_geostore_id method, so should maybe throw an exception if global.get_geostore_id() is called due to a bug.

app/models/pydantic/datamart.py
```diff
@@ -43,7 +99,7 @@ class DataMartResourceLinkResponse(Response):
 
 
 class TreeCoverLossByDriverIn(StrictBaseModel):
-    geostore_id: UUID
+    aoi: Union[GeostoreAreaOfInterest, AdminAreaOfInterest, Global] = Field(..., discriminator='type')
```
danscales, line 102: Not your change, but at line 90 above, the type should be Optional[DataMartMetaDdata], so it can have a value of None.

app/models/pydantic/datamart.py
```diff
@@ -1,12 +1,68 @@
...
+    @validator('provider', pre=True, always=True)
+    def set_provider_default(cls, v):
+        return v or 'gadm'
+
+    @validator('version', pre=True, always=True)
+    def set_version_default(cls, v):
+        return v or '4.1'
+
+
+class Global(AreaOfInterest):
```
danscales, line 64: Nit: This seems very generic and possibly confusing to call it 'Global'.  Feel free to completely ignore, but maybe rename it GlobalArea or GlobalAOI ?

danscales (commented):
Looks good.  Just some minor suggestions.

app/models/pydantic/datamart.py
solomon-negusse: I went with that since users will be specify it as either `aoi[global]` or  `{'aoi': 'type': 'global'} so it may be clear enough as a whole. Another issue is that not all datasets will have global extent, and I considered "full_extent" but went with global for simplicity. @jterry64: any strong opinion on this?

app/models/pydantic/datamart.py
solomon-negusse: I removed the `abstractmethod` since this isn't required for child classes anymore. As a compromise I have the base method return 'not implemented' error so child classes that don't implement it won't have to worry about it.

app/models/pydantic/datamart.py
solomon-negusse: Added some doc https://github.com/wri/gfw-data-api/pull/658/commits/2cf353654c4b2c5a0429eaf1492d3b58bac3211a

gtempus (approved):
The latest _GADM support_ changes aren't in here, but they shouldn't change this PR much.

jterry64:
@solomon-negusse when you get the chance could you resolve the merge conflicts? It's difficult to tell now which changes are youts

solomon-negusse:
> @solomon-negusse when you get the chance could you resolve the merge conflicts? It's difficult to tell now which changes are youts

All updated, @jterry64.

jterry64 (approved):

solomon-negusse:
@jterry64 i have added AOI metadata in c26ce1bca6f45077d1f8737953e10571e1ca7193

jterry64 (approved):
Sorry didn't see your message @solomon-negusse, the metadata changes look good to me!

PR #662, opened 2025-03-26
Return 409 for POST on existing resource, don't convert UUID to UUID

jterry64:
Fixes for two issues I found while testing in staging that return internal server errors.
- Using admin aoi causes an exception on `return UUID(geostore_id)` in `AdminAreaOfInterest` because `geostore_id` is already a UUID. I think this is correct, it's defined in our DB as UUID so it gets automatically converted to a UUID in Python. Wasn't quite sure how to add that to the tests though, because best I could do was fake a UUID response from DB, which doesn't tell us much.
- If you POST on an existing resource, it's just straight to the CRUD function and give a unique violation exception. I added a check for the resource first, and return a 409 Conflict if it exists. Alternatively, we could return a 200 OK with a link to the resource, thoughts?
- Refactored to _check_resource_exists function to make a bit more readable.

### DIFF
--- a/app/models/pydantic/datamart.py
+++ b/app/models/pydantic/datamart.py
@@ -51,7 +51,7 @@ class AdminAreaOfInterest(AreaOfInterest):
             region_id=self.region,
             subregion_id=self.subregion,
         )
-        return UUID(geostore_id)
+        return geostore_id
 
     @root_validator
     def check_region_subregion(cls, values):
--- a/app/routes/datamart/land.py
+++ b/app/routes/datamart/land.py
@@ -117,7 +117,12 @@ async def tree_cover_loss_by_driver_search(
     )
 
     # check if it exists
-    await _get_resource(resource_id)
+    resource_exists = await _check_resource_exists(resource_id)
+    if not resource_exists:
+        raise HTTPException(
+            status_code=404, detail="Resource not found, may require computation."
+        )
+
     link = DataMartResourceLink(
         link=f"{API_URL}/v0/land/tree_cover_loss_by_driver/{resource_id}"
     )
@@ -188,6 +193,13 @@ async def tree_cover_loss_by_driver_post(
         dataset_version,
     )
 
+    resource_exists = await _check_resource_exists(resource_id)
+    if resource_exists:
+        raise HTTPException(
+            status_code=409,
+            detail=f"Resource f{resource_id} already exists with those parameters.",
+        )
+
     await _save_pending_resource(resource_id, request.url.path, api_key)
 
     background_tasks.add_task(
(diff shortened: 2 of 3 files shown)


gtempus (approved):
Doh! Sorry about the `UUID` bug. 
Good catch on the `409 Conflict`

gtempus:
I think if we're going to refuse to do work on the `POST` then we should return the `409`

solomon-negusse (commented):
@jterry64 this SO post talks about how to make `POST` idempotent by returning 303 redirect to the existing resource when a duplicate request comes in. Given the resource having the same id in our case, this may suggest 201 is ok. But reading more, 409 seems like the standard response code. 

https://stackoverflow.com/questions/33249708/can-a-restful-post-method-be-implemented-to-be-idempotent

solomon-negusse:
More work on the client, but this [article](https://apisyouwonthate.com/blog/idemptoency-keys/) discusses `Idempotency-key` request header that ensures the request is unique (like our resourceId) and gives examples of APIs that implement that (most of them returning success and one that returns 409) based on that key.

jterry64:
> @jterry64 this SO post talks about how to make `POST` idempotent by returning 303 redirect to the existing resource when a duplicate request comes in. Given the resource having the same id in our case, this may suggest 201 is ok. But reading more, 409 seems like the standard response code.
> 
> https://stackoverflow.com/questions/33249708/can-a-restful-post-method-be-implemented-to-be-idempotent

@solomon-negusse from looking at how [Stripe API implement this](https://docs.stripe.com/api/idempotent_requests) (the one quoted there), I think we basically already do this by immediately creating a pending entry without actually doing the computation. So I don't think we need the idempotency key if I understand correctly? But, I do see their point about just sending back the same 201 response instead of 409 to make it a little easier for the client, since it's idempotent anyway. @gtempus do you have thoughts?

solomon-negusse:
>@solomon-negusse from looking at how [Stripe API implement this](https://docs.stripe.com/api/idempotent_requests) (the one quoted there), I think we basically already do this by immediately creating a pending entry without actually doing the computation. So I don't think we need the idempotency key if I understand correctly? But, I do see their point about just sending back the same 201 response instead of 409 to make it a little easier for the client, since it's idempotent anyway. @gtempus do you have thoughts?

I think so @jterry64. Also since we can guarantee the request's uniqueness using the resource id server side,  it seems to me that we can return 201 without needing the idempotency key (and without recomputing).  Would be good to hear from Gary.

jterry64:
@solomon-negusse I'm going to just merge this up and do a follow up PR if we want to change the status, I think we still have flexbility here while the frontend is implementing the basic widget

PR #664, opened 2025-04-01
Add DataMart docs to openapi, and clearly mark as beta

jterry64:
Providing this so the frontend can quickly can get up-to-date docs on the request/response structures and use swagger to form docs.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -242,7 +242,7 @@ def custom_openapi():
         {"name": "Task API", "tags": ["Tasks"]},
         {"name": "Analysis API", "tags": ["Analysis"]},
         {"name": "Health API", "tags": ["Health"]},
-        {"name": "Land API", "tags": ["Beta Land"]},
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -241,6 +242,7 @@ def custom_openapi():
         {"name": "Task API", "tags": ["Tasks"]},
         {"name": "Analysis API", "tags": ["Analysis"]},
         {"name": "Health API", "tags": ["Health"]},
+        {"name": "Land API", "tags": ["Beta Land"]},
```
danscales, line 245: Do you want to change this "Beta Land API"? Just asking, no need to change if that doesn't make sense.

danscales (approved):

app/main.py
jterry64: sure

jterry64: done

PR #672, opened 2025-04-09
Make static vector tile caches and raster tile sets use on-demand queue

jterry64:
To keep it safer, just starting with switching to on-demand the jobs we're currently being blocked by, since it does involve touching code in a quite few places to switch the queue. It looks like pre-commit also had a few things to say about these files, sorry for the mix of style updates.


solomon-negusse (approved):

dmannarino (approved):
Thank you!

PR #679, opened 2025-04-14
Download by AOI endpoint

jterry64:
Create download by AOI endpoint, re-using @gtempus's elegant AOI pydantic models from DataMart to allow for downloading by any AOI we support lookup for.

### DIFF
--- a/app/models/pydantic/datamart.py
+++ b/app/models/pydantic/datamart.py
@@ -5,6 +5,8 @@ from itertools import groupby
 from typing import Any, Dict, List, Literal, Optional, Union
 from uuid import UUID
 
+from fastapi import HTTPException, Request
+from jsonschema import ValidationError
 from pydantic import Field, root_validator, validator
 
 from app.models.pydantic.responses import Response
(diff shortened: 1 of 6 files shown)


app/routes/datasets/downloads.py
```diff
@@ -194,6 +197,78 @@ async def download_csv_post(
...
+    filename: str = Query("export.json", description="Name of export file."),
+    delimiter: Delimiters = Query(
+        Delimiters.comma, description="Delimiter to use for CSV file."
+    ),
+):
+    """Execute a READ-ONLY SQL query on the given dataset version (if
+    implemented) for a given AOI.
+
+    Returns downloads in either JSON or CSV depending on the Accept
+    header.
```
gtempus, line 259: I don't think this is true for these two endpoints. We've kept the format as part of the path.

app/routes/datasets/downloads.py
```diff
@@ -27,6 +30,46 @@
 
 router: APIRouter = APIRouter()
 
+OPENAPI_EXTRA_AOI = {
```
gtempus, line 33: For the datamart route, I put this in its `__init__.py` to clean things up a bit.

app/models/pydantic/datamart.py
```diff
@@ -202,3 +204,33 @@ def to_csv(
 
         csv_file.seek(0)
         return csv_file
+
+
+def _parse_area_of_interest(request: Request) -> AreaOfInterest:
```
gtempus, line 209: Should this still be prefixed with an underscore if we are importing into other modules?

app/routes/datasets/downloads.py
```diff
@@ -194,6 +237,80 @@ async def download_csv_post(
...
+    Returns downloads in either JSON or CSV depending on the Accept
+    header.
+    """
+
+    dataset, version = dataset_version
+
+    await _check_downloadability(dataset, version)
+
+    geostore_id = await aoi.get_geostore_id()
+    geostore: Optional[GeostoreCommon] = await get_geostore(geostore_id)
```
gtempus, line 302: Nice!

tests_v2/unit/app/routes/datasets/test_download.py
```diff
@@ -160,3 +180,75 @@ async def httpx_response_coroutine(*args, **kwargs) -> httpx.Response:
...
+        patch(
+            "app.routes.datasets.downloads.get_geostore", return_value=TEST_GEOSTORE
+        ) as mock_get_geostore,
+        patch(
+            "app.routes.datasets.downloads._query_dataset_csv",
+            return_value=StringIO("x,y\n,1,2"),
+        ) as mock_query_dataset_csv,
+    ):
+        response = await async_client.get(
+            f"/dataset/{dataset_name}/{version_name}/download_by_aoi/csv?sql={TEST_SQL}&aoi[type]=admin&aoi[country]=IDN&aoi[region]=9&aoi[subregion]=9",
```
gtempus, line 208: Just a heads up, I'm going to advocate that we pass in `aio[provider]` and `aoi[version]` when we build these URLs in the Subscriptions MS. Obviously, the endpoints should still work. I just want to be explicit with the gadm source going forward.

gtempus (changes requested):
Looks great, @jterry64. Thanks for adding a couple of tests, too! 💪

jterry64:
@gtempus thanks for the quick feedback! Addressed all your comments

gtempus (approved):

PR #682, opened 2025-04-18
Allow simplify download AOI

jterry64:
- Pass simplification to download by AOI
- Create geostore differently by AOI type to support unique parameters 
- Refactor to common method across CSV and JSON endpoints
- Make CSV endpoint use CSV extension filename by default
- Update GADM dataset version for prod
- Update tests to catch missed cases


gtempus (approved):
Looks good, @jterry64 ! 💯

PR #684, opened 2025-04-21
Fix TCL by driver DataMart response grouping and forward Accept header

jterry64:
- To properly group by the results from the OTF TCL by driver analysis, you need to first sort the results, since itertools.groupby() can only group consecutive keys.
- To keep consistent results order, I sort by the pixel value corresponding to each driver class. I don't think this totally matters other than for consistency, and I'm not the biggest fan of adding this logic to the code, so convince me I should remove this.
- Added additional years to the mocked results/resource to make sure tests catch this issue. Confirmed adding the years broke the tests before I added the fix.
- Add terraform changes to forward `Accept` header from CloudFront to the origin server. I believe this is why the tests pass but the `Accept` header seems to be ignored in production.

### DIFF
--- a/app/models/pydantic/datamart.py
+++ b/app/models/pydantic/datamart.py
@@ -145,13 +145,29 @@ class TreeCoverLossByDriverResult(StrictBaseModel):
             for row in rows
         ]
 
+        # sort rows first since groupby will only group consecutive keys
+        # this shouldn't matter, but to match existing sorting behavior, sort by
+        # mapped pixel value rather than alphabetical
+        driver_value_map = {
+            "Permanent agriculture": 1,
+            "Hard commodities": 2,
+            "Shifting cultivation": 3,
+            "Forest management": 4,
+            "Wildfires": 5,
+            "Settlements and infrastructure": 6,
+            "Other natural disturbances": 7,
+        }
+        sorted_rows = sorted(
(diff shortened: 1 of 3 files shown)


app/models/pydantic/datamart.py
```diff
@@ -145,13 +145,29 @@ def from_rows(rows):
...
+            key=lambda x: driver_value_map[x["tsc_tree_cover_loss_drivers__driver"]],
+        )
         tcl_by_driver = [
             {
                 "drivers_type": driver,
                 "loss_area_ha": sum([year["area__ha"] for year in years]),
             }
             for driver, years in groupby(
-                rows, key=lambda x: x["tsc_tree_cover_loss_drivers__driver"]
+                sorted_rows, key=lambda x: x["tsc_tree_cover_loss_drivers__driver"]
```
danscales, line 170: Nit: this is not in the current change, so feel free to ignore.  But I think the use of years here is hard to understand.  If I'm understanding correctly, years would be better named 'groupedrows' in line 169, and then could replace 

sum([year["area__ha"] for year in years])

with

sum([row["area__ha"] for row in groupedrows])

in line 167.

danscales (approved):
Seems good to me.

dmannarino (approved):

PR #692, opened 2025-05-02
Fix wdpa id column name 

solomon-negusse:


app/crud/geostore.py
```diff
@@ -288,7 +288,7 @@ async def get_wdpa_geostore_id(dataset, version, wdpa_id):
     sql: Select = (
         db.select(columns_etc)
         .select_from(src_table)
-        .where(db.text("wdpaid=:wdpa_id").bindparams(wdpa_id=wdpa_id))
+        .where(db.text("wdpa_pid=:wdpa_id").bindparams(wdpa_id=wdpa_id))
```
danscales, line 291: Just to be sure - you really mean wdpa_pid, here, rather that wdpa_id?  Is wdpa_pid the name of the field in the database?

danscales (approved):

app/crud/geostore.py
solomon-negusse: Yes, it's wdpa_pid in the database -- i just surfaced it using what I think is user friendly `wdpa_id`

PR #694, opened 2025-05-06
Add documentation for union_bands and band_count creation options (for the online doc)

danscales:
Add documentation for union_bands and band_count

Also, added descriptions for data_type and no_data.

I somewhat guessed on the meaning of band_count based on a few mentions in the code, so I definitely welcome comments on all of the descriptions, esp. union_bands and band_count.


dmannarino (dismissed):
Thanks for adding these!

PR #697, opened 2025-05-13
GTC-3293: Allow 1km drivers to be specified

dmannarino:
## Pull request type

- [x] Feature

If one were to specify the new drivers dataset as an override to the drivers datamart endpoint it would conflict with the old version.

Issue Number: GTC-3293

## What is the new behavior?

- If one specifies the new version of drivers as an override it replaces the old version, rather than conflicts with it.

- [ ] Yes
- [x] No

## Other information

Thanks to @gtempus for all the help!

### DIFF
--- a/app/routes/datamart/land.py
+++ b/app/routes/datamart/land.py
@@ -168,7 +168,17 @@ async def tree_cover_loss_by_driver_post(
     """Create new tree cover loss by drivers resource for a given geostore and
     canopy cover."""
 
-    dataset_version = DEFAULT_LAND_DATASET_VERSIONS | data.dataset_version
+    mutually_exclusive_datasets = {
+        "wri_google_tree_cover_loss_drivers": "tsc_tree_cover_loss_drivers"
+    }
+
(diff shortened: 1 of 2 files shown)


app/routes/datamart/land.py
```diff
@@ -273,3 +276,22 @@ def _get_metadata(
...
         sources=sources,
     )
+
+
+def apply_overrides_and_remove_mutually_exclusive_datasets(
+    dataset_versions: Dict[str, str]
+) -> Dict[str, str]:
+    """Given a dictionary of dataset:version pair overrides, return a
+    dictionary of default dataset versions with those overrides applied and
+    mutually exclusive datasets removed."""
```
gtempus, line 286: ❤️

gtempus (approved):
Nice work, @dmannarino ! 💪 
This wound up being a great exercise in TDD and the evolutionary design approach of "make the change easy, then make the change." 
It was fun pairing with you on this. 😃

PR #747, opened 2026-03-13
Add table of restricted dataset versions and deny access for those  queries

danscales:
Add table of restricted dataset versions and deny access for those  queries

This is analogous to the query permission check we already have for restricted datasets (which has only ever been
"wdpa_licensed_protected_areas"). I am just implementing a permission check for specific restricted dataset/versions as well.

I already added a restricted version umd_tree_cover_loss/v1.13, so that will prevent access to TCL2025 (until we remove the restriction), unless the query supplied a bearer token with ADMIN privileges for either gfw-pro or gfw-apps.

New test is test_query_restricted_version_disallowed, but I also made a re-factoring change to pull out common code from a bunch of the tests into a small helper function get_headers_with_origin.

### DIFF
--- a/app/authentication/token.py
+++ b/app/authentication/token.py
@@ -1,4 +1,4 @@
-from typing import cast
+from typing import cast, Tuple, List
 
 from fastapi import Depends, HTTPException
 from fastapi.logger import logger
@@ -6,8 +6,8 @@ from fastapi.security import OAuth2PasswordBearer
 from httpx import Response
 
 from ..models.pydantic.authentication import User
-from ..routes import dataset_dependency
-from ..settings.globals import PROTECTED_QUERY_DATASETS
+from ..routes import dataset_version_dependency
+from ..settings.globals import PROTECTED_QUERY_DATASETS, PROTECTED_QUERY_DATASET_VERSIONS
 from ..utils.rw_api import who_am_i
 
 # token dependency where we immediately cause an exception if there is no auth token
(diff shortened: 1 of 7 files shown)


tests_v2/conftest.py
```diff
@@ -300,6 +300,70 @@ async def generic_raster_version(
...
+@pytest_asyncio.fixture
+async def restricted_version(
+    async_client: AsyncClient,
+    restricted_dataset: Tuple[str, str],
+    monkeypatch: MonkeyPatch,
+) -> AsyncGenerator[Tuple[str, str, Dict[str, Any]], None]:
+    """Create licensed version."""
+
+    dataset_name, _ = restricted_dataset
+    version_name: str = "v1.13"
```
dmannarino: When we un-restrict TCL v1.13 this will break, which will slow down deploying when we want to release. Could you inject the dataset/version so that this doesn't break when we un-restrict this particular combination?

app/authentication/token.py
```diff
@@ -40,47 +40,48 @@ async def is_admin(token: str = Depends(oauth2_scheme)) -> bool:
...
     """
 
-    return await is_app_admin(token, "gfw", "Unauthorized")
+    return await is_app_admin(token, ["gfw"], "Unauthorized")
 
 
-async def is_gfwpro_admin_for_query(
-    dataset: str = Depends(dataset_dependency),
+async def is_admin_for_query(
+    dv: Tuple[str, str] = Depends(dataset_version_dependency),
```
dmannarino, line 47: is_admin_for_query is better than is_gfwpro_admin_for_query, though I think it's still slightly misleading since non-admins can query non-restricted datasets. How about is_authorized_for_query?

dmannarino:
Code looks good! Just two suggested changes.

tests_v2/conftest.py
danscales: Good suggestion. Since we'll be removing v1.13 from the table when we release, I instead created a new permanently restricted version umd_tree_cover_loss/v1.1, so the test won't be affected when we removed umd_tree_cover_loss/v1.13 from the table.

app/authentication/token.py
danscales: Good suggestion, renamed the function to is_authorized_for_query(), as you said.

dmannarino (approved):
Looks good!

solomon-negusse (approved):
Looks good!

PR #764, opened 2026-08-22
Vendor grid code from PixETL in order to remove it as a dependency

dmannarino:
At some point a few tests were added that called PixETL code directly, necessitating adding it to the deps. The problem is that then the resolved dependencies must accommodate BOTH the Data API AND PixETL. i.e. they become locked together, and it has become very difficult to upgrade the deps of either. But the test really uses just a little code, that for constructing grids. I finally decided that the benefit of being able to unlock dep upgrades outweighed the potential for this code to drift, so I "vendored" the relevant code (i.e. copied it into a Data API module) and moved on with my life. This is basically required for numerous other upgrades and improvements.


danscales:
Is there any issue with the failed tests, or are the tests being flaky?

dmannarino:
> Is there any issue with the failed tests, or are the tests being flaky?

Failing tests were unrelated to these changes, and the underlying issue has been fixed. :)

dmannarino:
Closing in favor of https://github.com/wri/gfw-data-api/pull/781

PR #765, opened 2026-08-22
Fix minor issues preventing use of numpy >= 2

dmannarino:
When upgrading to Numpy 2 all seems fine at first. When calling the raster analysis lambda (via the query route), however, one gets a cryptic 500 back. Apparently Numpy 2 is more strict about promoting data types. I will try to reproduce the exact error this fixes and add it here.


danscales:
Can you put in the description the numpy complaint that you fixed. That would help make it clearer why you need to make the change. Or maybe a comment in the code about what the type of A is and why you need to convert to uint16?

dmannarino:
Closing for now since I can't seem to reproduce the error

PR #766, opened 2026-08-22
Fix off-by-one bug in dateconf formula

dmannarino:
I'm pretty sure this fixes a bug in the code. Say the first value in A is 0, which should be 'no data', if we add it to datetime64('2015-01-01', 'D') all of a sudden we have an alert on Jan 1st. This probably hasn't manifested much because of the symmetry of the encode/decode functions or filtering done elsewhere, but it is incorrect.


danscales (approved):
I'm not sure that no_data pixels will really get converted to an alert - it is probably filtered out in functions using DerivedLayer, right? Else no_data values would always become some date. But I do think this is probably a bug, where the raster SQL special date accessor for date_conf implementations (e.g. wur_radd_alerts__date) would decode with an off-by-one date, so the filtering would be wrong by 1 day.

So, I think this is the right fix to correct it, but maybe @jterry64  can take a look too.

PR #781, opened 2026-09-09
Mock PixETL functions in tests and remove it as a dependency

dmannarino:
At some point I added a few tests for the resample script, which in Batch makes a small use of the PixETL code.
This necessitated adding it to the test dependencies, which was done. The problem is that then the resolved dependencies must accommodate BOTH the Data API AND PixETL. i.e. they become locked together, and it has become very difficult to upgrade either. But the resample script really uses just a little PixETL code, that for constructing grids.
In a previous PR I tried vendoring the grid code from PixETL but got no approvals and hated it anyway. This PR instead mocks the called PixETL functions which I like a lot more as a solution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,12 +52,7 @@ dependencies = [
 dev = [
     "asgi-lifespan",
     "docker",
-    "geopandas", # Needed by pixetl in batch script test
-    # Pixetl is already installed in the pixetl image that's run in Batch,
-    # this is to enable tests to run in the test container:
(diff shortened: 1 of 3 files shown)


danscales (approved):
Looks fine, so I'll approve. It is weirdly saying one of six checks is missing: "build Expected — Waiting for status to be reported", but I do see that your tests succeeded on Sept 8th here:https://github.com/wri/gfw-data-api/actions/runs/34312015892

